PR #21, opened 2020-06-01
Add ASDF support

bkuehnert:
Needs to be tested further, as it sometimes throws an error
when compiling case-preservation.lisp. After restarting the
REPL, it works fine.


tests/cl-protobufs-tests.asd
```diff
@@ -39,6 +39,7 @@
    ;;   text-format-test.lisp
    ;;   zigzag-test.lisp
 
+   #+protobuf-file-debugged
```
Slids, line 42: do we need this?

tests/cl-protobufs-tests.asd
```diff
@@ -78,11 +79,11 @@
     :components ((:protobuf-file "extend-test.proto")
                  (:file "extend-test")))
 
-   #+protobuf-file-debugged
+;   #+protobuf-file-debugged
```
Slids, line 82: just remove

PR #47, opened 2020-06-08
Fix import test

bkuehnert:
* Adds an additional test to import-tests which checks that we are able to make the messages that are imported.
* Changes import-test to verify that we are able to load .proto files with import statements when the .proto files are specified in the correct order (with respect to imports). ASDF will not be able to load the .proto files unless the files are specified in the correct order.

### DIFF
--- a/tests/cl-protobufs-tests.asd
+++ b/tests/cl-protobufs-tests.asd
@@ -163,9 +163,9 @@
    (:module "import-test"
     :serial t
     :pathname ""
-    :components ((:protobuf-source-file "import-proto")
-		 (:protobuf-source-file "import-test-import-1")
+    :components ((:protobuf-source-file "import-test-import-1")
 		 (:protobuf-source-file "import-test-import-2")
+		 (:protobuf-source-file "import-proto")
 		 (:file "import-test")))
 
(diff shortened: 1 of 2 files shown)


tests/cl-protobufs-tests.asd
```diff
@@ -163,9 +163,9 @@
    (:module "import-test"
     :serial t
     :pathname ""
-    :components ((:protobuf-source-file "import-proto")
-		 (:protobuf-source-file "import-test-import-1")
+    :components ((:protobuf-source-file "import-test-import-1")
 		 (:protobuf-source-file "import-test-import-2")
+		 (:protobuf-source-file "import-proto")
```
bkuehnert, line 168: Although the formatting looks wrong here, it is correct on my branch. Not sure what to make of that. https://github.com/bkuehnert/cl-protobufs/blob/932d1d4cac9dc6d75f8e6f80482e3201dcc64663/tests/cl-protobufs-tests.asd#L168

Slids: Tab v space formatting thing? 
No idea...

Slids: Yea, thats it.

PR #49, opened 2020-06-09
Fix example command

bkuehnert:
Since the command is run from the directory containing cl-protobufs
this command will fail, as case-preservation.proto is not present
here. Thanks to dougk for spotting this.


bkuehnert:
All tests pass on my computer, the fact that tests fail on on the github actions seems to be a problem with github, rather than the PR.

Slids:
Adding internally due to annoying reasons...

PR #50, opened 2020-06-10
Implement optimized group serialization

bkuehnert:
*make-serializer works for non-repeated group fields
*make-serializer works for repeated group fields
*added an additional message which tests this functionality.
(also serves as a test to see if the tests correctly pass in the CI)


tests/serialization-tests.lisp
```diff
@@ -95,21 +99,32 @@ Parameters:
          (test6  (proto-impl:make-object basic-test6
                                          :intvals '(2 3 5 7)
                                          :strvals '("two" "three" "five" "seven")
-                                         :recvals (list test2 test2b))))
+                                         :recvals (list test2 test2b)))
+         (group1 (proto-impl:make-object subgroups
```
Slids, line 103: (make-object subgroups ...

bkuehnert: I'll make this change (and do so for the rest of the file) in a separate PR. This is because I have another branch for implementing deserialization, and it seems simpler to merge the deserialization and serialization branches before merging in this change.

Slids: sgtm

tests/serialization.proto
```diff
@@ -70,6 +70,15 @@ message BasicTest6 {
   repeated BasicTest2 recvals = 3;
 }
 
+message BasicTest7 {
+  required string strval = 1;
```
Slids, line 74: Please remove,
new required fields are not allowed internally.

PR #52, opened 2020-06-11
Implement optimized group deserialization

bkuehnert:
This PR allows make-deserializer to handle messages which have group fields. 
It also includes the repeated group tests added in https://github.com/qitab/cl-protobufs/pull/50.


serialize.lisp
```diff
@@ -783,10 +783,10 @@ Parameters:
            (ret-list (tags collectors)
              (return-from generate-field-deserializer (values tags collectors
                                                               nslot rslot)))
-           (call-pseudo-method (fun msg vbuf start end)
+           (call-pseudo-method (fun msg vbuf start end &optional (end-tag 0))
```
Slids, line 786: should we change this function name?
maybe flet-bound-call-pseudo-method (or something better)
so nobody gets confused again?

tests/serialization.proto
```diff
@@ -70,6 +70,15 @@ message BasicTest6 {
   repeated BasicTest2 recvals = 3;
 }
 
+message BasicTest7 {
+  required string strval = 1;
+  repeated int32 intvals = 2;
+  repeated group Subgroups = 3 {
+    required string strval = 1;
+    repeated int32 intvals = 2;
+  }
```
Slids, line 79: could you add a final field, maybe optional string?
This will add a test that it can actually read the element after
the (repeated) group(s)

bkuehnert: Good call, done.

serialize.lisp
bkuehnert: Maybe, I'm not even sure why it's redefined here anyway. I guess its for a debugging mode?

Also, looking at it again, it doesn't even make sense to call it "call-pseudo-method" at all. The parameters only make sense when the pseudo-method in question just so happens to be :deserialize (or :serialize, but that should never happen).

I think we either 
1. Remove it 
2. Rename it to something like call-deserializer and remove the "fun" argument such that its always set to :deserializer.
3. Replace "msg vbuf start end" with "&rest args" so that it actually mirrors how call-pseudo-method is defined, and keep the name the same. 

which sounds the most reasonable?

Slids: 3 is never reasonable. If possible you should always write the parameters.
Probably 2.
It has an if statement which is used.

Slids: Never is too extreme, almost never.

bkuehnert: I think this a case in which it's not possible to write the parameters, since the lambda list for call-pseudo-method is just (fun &rest args). Anyway, I'll go with option 2. After looking at it again it seems like the best choice.

PR #63, opened 2020-06-23
Make emit-field definition global

bkuehnert:
The emit-field function is necessary for map types, as we need
to be serialize arbitrary types which aren't necessarily wrapped
in field descriptors/objects.

This also moves the treatment of non repeated fields to its own
function, which is what the map serialization subroutine will use 
(as map key/value types are never repeated).

### DIFF
--- a/pkgdcl.lisp
+++ b/pkgdcl.lisp
@@ -120,6 +120,10 @@
 
   (:shadow
    #:find-method)
+
+  (:import-from :alexandria
+                #:if-let)
+
   (:import-from :closer-mop
                 #:class-slots
                 #:class-direct-slots
--- a/pkgdcl.lisp
+++ b/pkgdcl.lisp
@@ -120,10 +120,6 @@
 
   (:shadow
    #:find-method)
-
-  (:import-from :alexandria
-                #:if-let)
-
   (:import-from :closer-mop
                 #:class-slots
                 #:class-direct-slots
--- a/pkgdcl.lisp
+++ b/pkgdcl.lisp
@@ -80,6 +80,7 @@
    #:serialize-object-to-stream
    #:serialize-object-to-bytes
    #:serialize-object
+   #:emit-non-repeated-field
    #:deserialize-object-from-file
    #:deserialize-object-from-stream
    #:deserialize-object-from-bytes
--- a/pkgdcl.lisp
+++ b/pkgdcl.lisp
@@ -80,7 +80,6 @@
    #:serialize-object-to-stream
    #:serialize-object-to-bytes
    #:serialize-object
-   #:emit-non-repeated-field
    #:deserialize-object-from-file
    #:deserialize-object-from-stream
    #:deserialize-object-from-bytes
--- a/serialize.lisp
+++ b/serialize.lisp
@@ -157,9 +157,8 @@ Parameters:
                                    ;; unnecessarily.
                                    (and (not (proto-lazy-p field))
                                         reader))
-                   (list object)
-                   type (proto-packed field) index buffer))
-
+                   (list object))
+               type (proto-packed field) index buffer)
               (undefined-field-type "While serializing ~S,"
                                     object type field))
           (or (emit-non-repeated-field
--- a/serialize.lisp
+++ b/serialize.lisp
@@ -122,6 +122,12 @@
       (iincf size (emit-field object field buffer)))))
 
 (defun emit-field (object field buffer)
+  "Serialize a single field from an object to buffer
+
+Parameters:
+  OBJECT: The protobuf object which contains the field to be serialized.
+  FIELD: The field of the object to serialize.
+  BUFFER: The buffer to serialize to."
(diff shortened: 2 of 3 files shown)


bkuehnert:
Changed to a draft, as I'm currently not sure if this is the best solution.

pkgdcl.lisp
```diff
@@ -80,6 +80,7 @@
    #:serialize-object-to-stream
    #:serialize-object-to-bytes
    #:serialize-object
+   #:emit-non-repeated-field
```
Slids, line 83: should definitely not be exported.
We really should remove most things from being exported...

bkuehnert: Ah ok. It seemed like almost everything that was used outside of file was being exported.

Slids:
Yes, that was a poor choice by whoever made it

serialize.lisp
```diff
@@ -117,161 +117,170 @@
...
-                        (undefined-field-type "While serializing ~S,"
-                                              object type field)))))))
-      (let ((size 0))
-        (dolist (field (proto-fields msg-desc) size)
-          (iincf size (emit-field object field)))))))
+  (let ((size 0))
+    (dolist (field (proto-fields msg-desc) size)
+      (iincf size (emit-field object field buffer)))))
+
+(macrolet ((read-slot (object slot reader)
```
Slids, line 124: I do not like top level macrolets, I prefer to call them macros.
moreso why is this a macrolet? 
Probably an attempt to force inline?
Make it a label?

bkuehnert: Changed it to a flet. There's no difference in performance

serialize.lisp
```diff
@@ -117,161 +117,170 @@
...
-                        (undefined-field-type "While serializing ~S,"
-                                              object type field)))))))
-      (let ((size 0))
-        (dolist (field (proto-fields msg-desc) size)
-          (iincf size (emit-field object field)))))))
+  (let ((size 0))
+    (dolist (field (proto-fields msg-desc) size)
+      (iincf size (emit-field object field buffer)))))
+
+(defun emit-field (object field buffer)
```
Slids, line 124: There's emit-field and emit-non-repeated field, I'm wondering if we should have an emit-field and an emit-repeated-field and maybe we can win by having the emit-repeated-field call emit field.
I think at least the duality is useful.

serialize.lisp
```diff
@@ -1,4 +1,4 @@
-;;; Copyright 2012-2020 Google LLC
+;; Copyright 2012-2020 Google LLC
```
Slids, line 1: missed a ;

cgay:
Hi Ben, could you fix up these lint warnings that were generated internally?:
```
No docstring for function EMIT-FIELD. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/serialize.lisp:123
  Lines should not end in whitespace. (WHITESPACE-AT-EOL)
	//depot/google3/third_party/lisp/cl_protobufs/serialize.lisp:240
```
Thanks

cgay:
Wow, fixed in < 28 seconds.

bkuehnert:
> Wow, fixed in < 28 seconds.

serialize.lisp
```diff
@@ -117,161 +117,191 @@
...
+            (has-extension object (slot-value field 'internal-field-name))
+            (= (bit (slot-value object '%%is-set)
+                    (slot-value field 'field-offset))
+               1))
+      (return-from emit-field 0))
+    (let ((type   (slot-value field 'class))
+          (slot   (slot-value field 'internal-field-name))
+          (reader (slot-value field 'reader))
+          (index (proto-index field))
+          msg)
```
Slids, line 150: remove

serialize.lisp
```diff
@@ -117,161 +117,191 @@
...
+
+Parameters:
+  VALUE: The data to serialize, e.g. the data resulting from calling read-slot on a field.
+  TYPE: The :class slot of the field.
+  PACKED-P: Whether or not the field in question is packed.
+  INDEX: The index of the field (used for making tags).
+  BUFFER: The buffer to write to."
+  (let ((size 0)
+        msg)
+    (declare (fixnum size index))
```
Slids, line 180: outside the let

serialize.lisp
```diff
@@ -117,161 +117,191 @@
...
+    (declare (buffer buffer))
```
Slids, line 181: outside the let

serialize.lisp
```diff
@@ -117,161 +117,191 @@
...
+
+Parameters:
+  VALUE: The data to serialize, e.g. the data resulting from calling read-slot on a field.
+  TYPE: The :class slot of the field.
+  INDEX: The index of the field (used for making tags).
+  BUFFER: The buffer to write to."
+  (let ((size 0)
+        msg)
+    (declare (fixnum size index))
+    (declare (buffer buffer))
```
Slids, line 255: move above let

PR #65, opened 2020-06-23
Fix bug with group serialization

bkuehnert:
The optimized serializer would compute a smaller than actual
length for protos which contained nested messages which had group
fields (Not sure why someone would ever do this, but it now at
least works).

### DIFF
--- a/serialize.lisp
+++ b/serialize.lisp
@@ -700,7 +700,7 @@ See field-descriptor for the distinction between index, offset, and bool-number.
                          `(let ((,vval ,reader))
                             (when ,vval
                               (iincf ,size (encode-uint32 ,tag1 ,vbuf))
-                              ,(call-pseudo-method :serialize msg vval vbuf)
+                              (iincf ,size ,(call-pseudo-method :serialize msg vval vbuf))
                               (iincf ,size (encode-uint32 ,tag2 ,vbuf)))))
                        (let ((tag (make-tag $wire-type-string index)))
                          `(let ((,vval ,reader))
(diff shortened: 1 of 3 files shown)


serialize.lisp
```diff
@@ -656,7 +656,7 @@ See field-descriptor for the distinction between index, offset, and bool-number.
                            `(when ,boundp
                               (,iterator (,vval ,reader)
                                          (iincf ,size (encode-uint32 ,tag1 ,vbuf))
-                                         ,(call-pseudo-method :serialize msg vval vbuf)
+                                         (iincf ,size ,(call-pseudo-method :serialize msg vval vbuf))
```
Slids, line 659: new line

cgay:
Ben, could you add a test please? Thanks.

serialize.lisp
cgay: The internal lint rules use 100-column max.

bkuehnert:
> Ben, could you add a test please? Thanks.

Done. Writing the test exposed the fact that I neglected to fix the bug for non-repeated groups, so that's good.

cgay:
woo!  Sorry to be annoying, but could you remove all tabs from the PR. They won't play nice with diffing and such.  Best if you set your editor to never use literal tabs. Thanks.

bkuehnert:
> woo! Sorry to be annoying, but could you remove all tabs from the PR. They won't play nice with diffing and such. Best if you set your editor to never use literal tabs. Thanks.

Huh, I've had this problem before and have explicitly told emacs to never use tabs. Thanks for catching it, time to fix my emacs config.

Slids:
Been, please fix the line size problem.
The linter won't let us submit until you do...
#Ihatelintererrorsthatwontsubmit

bkuehnert:
Whoops. I thought "100 char limit" meant 100 inclusive. Should be good now.

PR #67, opened 2020-06-26
Add map type

bkuehnert:
Currently is a draft since there are still a few problems I need to work through, but I wanted to start the review process now.

Some problems:
* I lean on this helper function proto-type->keyword in a few places, which makes the following conversion: primitive lisp types (such as proto:int32) map to keywords (so :int32), lisp types that describe messages, (so (or null basic-test2) for example) get mapped to their underlying message type (so basic-test2 in that example), and lisp types that describe enums are left alone. So, in a sense it's just translating from the proto-set-type to the proto-class on the field descriptor. It seems like these two things basically store the same data, so one solution to this problem is to just remove one. As writing this out, I think I should at the very least rename the helper function to set-type->class.
* Part of what makes the above hack so ugly is the (or null basic-test2) -> basic-test2 portion of the function. It would be a lot better if types corresponding to messages never had the (or null ... ) wrapper in the first place. This wrapper is put on by protoc, and I've tried adding a clause that doesn't do this if the type in question is being generated in a map field. I got a strange bug where it works when I add a print statement, and stops working when I remove that print statement. I've tried running make-clean in between compilations and it doesn't fix it. 
* The tests currently fail, since proto-equal isn't able to consistently tell if two hash-tables are equivalent (under the hood it's calling equalp, which should work, so this is confusing). Note: This problem mysteriously disappeared, and I can't find which commit fixed it. 
* Documentation in the README, more comments where appropriate
* Printing does not sort by key value, which is required by the protobuf documentation.
* There aren't enough tests


define-proto.lisp
```diff
@@ -601,7 +696,13 @@ Arguments:
                `(,hidden-accessor-name ,obj)))
 
         ,@(make-common-forms-for-structure-class
-           proto-type public-slot-name slot-name field)))))
+           proto-type public-slot-name slot-name field)
+
+
+        ; Make special map forms.
+        ,@(if (eq (proto-class field) :map)
```
Slids, line 703: when

define-proto.lisp
```diff
@@ -409,6 +410,47 @@ Parameters:
...
+                  :index index
+                  :internal-field-name internal-slot-name
+                  :external-field-name slot
+                  :reader reader)))
+    `(progn
+       define-map
+       nil
+       nil
+       ,mfield
+       ,mslot)))
```
Slids, line 452: please add comments as to what each of these fields are.
The return thing should be a &key clause and these hsould be :stuffs
but tech debt will tech debt

protoc/message.cc
```diff
@@ -68,112 +68,115 @@ void MessageGenerator::GenerateSource(io::Printer* printer,
...
-
-  if (descriptor_->enum_type_count() > 0) {
-    printer->Print("\n;; Nested enums.");
-    for (int i = 0; i < descriptor_->enum_type_count(); ++i) {
-      enums_[i]->Generate(printer);
+  // If descriptor_ describes a map entry message, ignore it. We do not use
+  // this generated message for map types.
+  if (descriptor_->options().map_entry()) {
+    return;
+  } else {
```
Slids, line 75: remove the else, no need

serialize.lisp
```diff
@@ -1,4 +1,4 @@
-;;; Copyright 2012-2020 Google LLC
+;; Copyright 2012-2020 Google LLC
```
Slids, line 1: remoeve a ;

serialize.lisp
```diff
@@ -117,161 +117,196 @@
...
-                              ;; To serialize a group, we encode a start tag,
-                              ;; serialize the fields, then encode an end tag
-                              (let ((tag1 (make-tag $wire-type-start-group index))
-                                    (tag2 (make-tag $wire-type-end-group   index)))
-                                (iincf size (encode-uint32 tag1 buffer))
+  (let ((size 0))
+    (dolist (field (proto-fields msg-desc) size)
+      (iincf size (emit-field object field buffer)))))
+
+(macrolet ((read-slot (object slot reader)
```
Slids, line 124: indenting is definitely off

Slids, line 124: top level macrolet?
Why not a defmacro

utilities.lisp
```diff
@@ -259,23 +259,34 @@ Arguments
...
-        (intern (format nil "~a.~a"
-                        (symbol-name proto-type)
-                        (symbol-name slot))
-                (symbol-package proto-type)))))
+                    (:get nil)
+                    (:map-get 'get)
+                    (:map-put 'put)
+                    (:map-rem 'remove))))
+    (cond ((or (eq f-symbol 'get) (eq f-symbol 'put) (eq f-symbol 'remove))
+           (intern (format nil "~a.~a-~a"
```
cgay, line 274: [pre-existing]
`~a` obeys `cl:*print-case*` so unless you want this to generate symbols like `|map-test.my-map-get|` you need to use `(intern (nstring-upcase (format ...)))`

utilities.lisp
```diff
@@ -377,6 +395,29 @@ Arguments:
...
 
+;; Due to the fact that serialization works with keywords for primitive
+;; types, we need this hack for now. If the type given is not primitive,
+;; then it is some message type of the form (or null [type]). In which
+;; case, return [type].
+(defun proto-type->keyword (type)
+  (if (and (listp type)
+           (eq (first type) 'or)
+           (eq (second type) 'null))
+      (proto-type->keyword (third type))
```
cgay, line 406: It might be the case that currently it's not possible for TYPE to be `(or null x y)` or `(or x null)` but we shouldn't write the code to assume that. Future Me might inexplicably change (or null x) to (or x null) and have a heck of a time tracking down the resulting problems. How about `(if (nullable-p type) (non-null-subtype type) ...)`?

Again can be a different PR.

utilities.lisp
```diff
@@ -259,23 +259,34 @@ Arguments
...
-                (symbol-package proto-type))
-        (intern (format nil "~a.~a"
-                        (symbol-name proto-type)
-                        (symbol-name slot))
-                (symbol-package proto-type)))))
+                    (:get nil)
+                    (:map-get 'get)
+                    (:map-put 'put)
+                    (:map-rem 'remove))))
+    (cond ((or (eq f-symbol 'get) (eq f-symbol 'put) (eq f-symbol 'remove))
```
cgay, line 273: `(cond ((member f-symbol '(get put remove)) ...`

serialize.lisp
```diff
@@ -117,161 +117,196 @@
 ;; The default method uses metadata from the message descriptor.
 (defmethod serialize-object (object (msg-desc message-descriptor) buffer)
   (declare (buffer buffer))
-  (macrolet ((read-slot (object slot reader)
```
cgay, line 120: Why such a big diff here?  If it's mostly a whitespace change please do that in a separate PR. I can't tell how much of this needs reviewing. (Would be better using the internal diff tools but alas...)

cgay:
The new interfaces will need to be documented, but that can be in a separate PR unless Jon wants it in the same one.

define-proto.lisp
```diff
@@ -409,6 +410,47 @@ Parameters:
...
+(defmacro define-map (name (&key type index))
+"Define a lisp type given the data for a protobuf map type.
+
+Parameters:
+  NAME: Map type name
+  TYPE: The proto type of the map, in the form of (proto:map-of [key-type] [val-type])
+  INDEX: Index of this map type in the field."
+  (check-type index integer)
+  ; binding slot to this seems incorrect, but changing it breaks.
+  (let* ((slot      (or name (and name (proto->slot-name name *package*))))
```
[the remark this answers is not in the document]
Slids: More is true!
These macros were created with the idea that someone might write the lisp schema by hand.
I would consider this a bug, so we should probaly go through all of these define forms and fix them given anyone who writes these by hand is wrong.

define-proto.lisp
```diff
@@ -537,6 +579,59 @@ Arguments:
 
         (export '(,has-function-name ,clear-function-name ,public-accessor-name))))))
 
+(defun make-map-structure-class-forms (proto-type public-slot-name slot-name field)
+  "Create specific forms needed for map fields: set, get, and remove.
```
Slids, line 583: explain how these are different from the usual ones

define-proto.lisp
```diff
@@ -537,6 +579,59 @@ Arguments:
...
+Arguments:
+  PROTO-TYPE: The Lisp type name of the proto message.
+  PUBLIC-SLOT-NAME: Public slot name for the field (without the #\% prefix).
+  SLOT-NAME: Slot name for the field (with the #\% prefix).
+  FIELD: The class object field definition of the field."
+  (let ((public-getter-name (proto-slot-function-name proto-type public-slot-name :map-get))
+        (public-setter-name (proto-slot-function-name proto-type public-slot-name :map-put))
+        (public-remove-name (proto-slot-function-name proto-type public-slot-name :map-rem))
+        (hidden-accessor-name (fintern "~A-~A"  proto-type slot-name))
+        (val-default-form (get-default-form (third (proto-set-type field))
```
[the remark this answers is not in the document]
Slids: For now at least take it in the let call.
Don't do this over and over.

define-proto.lisp
```diff
@@ -537,6 +579,59 @@ Arguments:
...
+                                            $empty-default))
+        (is-set-accessor (fintern "~A-%%IS-SET" proto-type))
+        (index (proto-field-offset field)))
+
+    (with-gensyms (obj new-key new-val)
+      `(
+        (declaim (inline ,public-getter-name))
+        (defun ,public-getter-name (,obj ,new-key)
+          (check-type ,new-key ,(second (proto-set-type field)))
```
Slids, line 603: instead use 
(declaim (type ,(second (proto-set-type field)) ,new-key)
Good write up:
https://stackoverflow.com/questions/60104101/common-lisp-difference-between-declare-check-type

define-proto.lisp
```diff
@@ -654,7 +755,9 @@ Arguments:
               (eq (first type) 'cl:or)
               (eq (second type) 'cl:null))
```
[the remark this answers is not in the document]
Slids: Ah, this is something I can comment on.
You have to look at the output of the protoc.
For example:
in type.proto->type.lisp
(source-context  :index 5  :type (cl:or cl:null cl-protobufs.google.protobuf::source-context) 
the first element is an or and the second is a null.
We can probably now just check that the first is an or given
some update I made with proto structs adding a type.
Ben can you verify?

define-proto.lisp
```diff
@@ -654,7 +755,9 @@ Arguments:
...
          nil)
-        ;; The only possibility left is we have an enum.
+        ((and (listp type)
+              (eq (first type) 'proto:map-of))
+         '(make-hash-table))
         ((enum-default-value `,type) (enum-default-value `,type))
         (t `(enum-default-value ',type))))))
```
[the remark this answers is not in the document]
Slids: Because we may have not yet evaluated the define-enum.
This is a bug I've yet to fix.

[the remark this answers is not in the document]
Slids: Ben could you add a comment explaining this.

text-format.lisp
```diff
@@ -126,11 +126,12 @@ Parameters:
             (format stream "~&}~%"))
         nil))))
 
-(defun print-prim (val type field stream indent)
+(defun print-prim (val type field stream indent &optional (newline-p t))
```
Slids, line 129: why?

text-format.lisp
```diff
@@ -126,11 +126,12 @@ Parameters:
...
   (when (or val (eq type :bool))
-    (if (eq indent 't)
-      (format stream "~A: " (proto-name field))
-      (format stream "~&~VT~A: " (+ indent 2) (proto-name field)))
+    (unless (eq indent 't)
```
Slids, line 131: when not unless indent?
Either it's (not nil) or it's nil, 
non of this eq t.

text-format.lisp
```diff
@@ -151,9 +166,9 @@ Parameters:
          (format stream "\"~A\"" val)))
       ((:date :time :datetime :timestamp)
        (format stream "~D" val)))
-    (if (eq indent 't)
-      (format stream " ")
-      (format stream "~%"))))
+    (if (and (not (eq indent 't)) newline-p)
```
Slids, line 169: again, odn't do eq 't

text-format.lisp
Slids: Oh god I see this is the same as before,
I'm... sad... don't do this.

text-format.lisp
Slids: :(

serialize.lisp
Slids: Actually, this comes from a different pr where I made my suggestions so please submit that pr first.
As I said, this should probably be a labels or a flet INSIDE the function.
This has no reason to be a macrolet except for inlining, and you should do a perf check to make sure but I doubt any gain will be noticeable...
A tl macrolet makes since if you want to use it in multiple other tl forms like multiple defuns, but thats not the case. A macrolet may be useful for closing over something in the ambient space, but again, not the case.

serialize.lisp
```diff
@@ -505,32 +540,32 @@ See field-descriptor for the distinction between index, offset, and bool-number.
...
-                    (when inner-index
-                      (push inner-index offset-list)))
-                   ;; Otherwise we have to mark is set later.
-                   (t
-                    (progn
-                      (push (cadr cell) initargs-final)
-                      (push (car cell) initargs-final)
-                      (when inner-index
-                        (push inner-index offset-list)))))))
+            (let* ((field (car cell))
```
Slids, line 543: This white space makes the cl impossible to see real diffs, please remove the white space change.

serialize.lisp
```diff
@@ -565,69 +600,145 @@ See field-descriptor for the distinction between index, offset, and bool-number.
                    (lazy  (proto-lazy-p field))
                    (type (proto-class field))
                    (data))
+              ; CELL nows points to the cons where DATA should go
+              (cond
```
Slids, line 604: so there's a lot of repeating going on here. 
You should be able to reformulated this so you don't repeat the below code.
We can talk about this more.
Also my pixelbook and github makes this diff impossible to read...

serialize.lisp
```diff
@@ -637,78 +748,144 @@ See field-descriptor for the distinction between index, offset, and bool-number.
...
+             `(when ,boundp
+                (let ((,vval ,reader))
+                  (flet ((serialize-pair (k v)
+                           (let ((ret-len (encode-uint32 ,tag ,vbuf))
+                                 (map-len 0))
+                             (with-placeholder (,vbuf)
+                               (iincf map-len (serialize-prim k ,key-type
+                                                              ,(make-tag key-type 1)
+                                                              ,vbuf))
+                               ,(cond ((and (keywordp val-type) (not (eq val-type :map)))
```
Slids, line 822: same here.
Should never copy code, it's a bad smell.
Note: cl-protobufs is filled with it...
Have to take a closer look and I'm guessing that's what we'll be doing on monday.

text-format.lisp
cgay: [without having looked at the code] This is a thing I've seen people do in Lisp when they want a ternary value. You could imagine the possibilities being T, NIL, and a string in this case. (Yeah, gross.)

Slids: I call this possibility stupid, 
but I don't doubt it...

define-proto.lisp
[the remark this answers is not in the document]
bkuehnert: Right, if protoc outputs a type of the form (cl:or ...), it must be of the form (cl:or cl:null cl-protobufs.[message]).

model-classes.lisp
```diff
@@ -234,6 +234,15 @@ Parameters:
     (or (find-message type)
         (find-type-alias type))))
 
+(defvar *maps* (make-hash-table :test 'eq)
+  "Maps map names (symbols) to message-descriptor instances
+   which describe map entries.")
```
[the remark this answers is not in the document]
bkuehnert: Whoops, this should've been removed. Originally, the protoc plugin would generate a message like MapFieldEntry { key_type key = 1; value_type value = 2; } if you had a field of type map<key_type,value_type>. I used the message descriptor for MapFieldEntry as a meta-object which described the map.

define-proto.lisp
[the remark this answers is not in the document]
bkuehnert: I've moved it inside the let for this function, but I use the same accessors several times in serialize.lisp, I agree that a separate function would be good.

I'm thinking of making functions (proto-map-key-type field) and (proto-map-val-type field) which just call (second (proto-set-type field)) and (third (proto-set-type-field)) respectively. Would model-classes.lisp be appropriate for these functions?

define-proto.lisp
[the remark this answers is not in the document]
bkuehnert: This is currently wrong. Name should never be nil, since "name" here is in the slot that "type" is in the other macros. I'll fix this now.

serialize.lisp
bkuehnert: Marking as resolved since it'll be updated once the move-emit-field changes are merged to this branch

tests/map-test.lisp
```diff
@@ -0,0 +1,105 @@
+;;; Copyright 2020 Google LLC
+;;;
+;;; Use of this source code is governed by an MIT-style
+;;; license that can be found in the LICENSE file or at
+;;; https://opensource.org/licenses/MIT.
+
+(defpackage #:cl-protobufs.test.map-test
+  (:use #:cl
+        #:clunit
+        #:cl-protobufs.map-test
```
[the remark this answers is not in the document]
Slids: Require trivial-local-knicknames om the asdf file
Then just use (:local-nicknames (#:cl-protobufs.map-test #:map-test)
I've suggested this a number of times...

bkuehnert:
Update: 
1. Moved map data to a map-descriptor object (instead of keeping everything in proto-set-type). See reasoning in commit 0daa892.
2. Got rid of code duplication in optimized serializer and generic deserializer.

text-format.lisp
bkuehnert: This isn't necessary at all, and I'll change it once my other changes to this function get merged.

bkuehnert:
I've tried getting rid of the code duplication in generate-field-deserializer (lines 1047-1071 of serialize.lisp). I tried a few different ways to break the function into smaller pieces, but I couldn't find a good way. I think the code is just too different compared with the source of the duplication (lines 1079-1107 of serialize.lisp).

Slids:
What about:
 ,(cond ((keywordp val-class)
                                     `(multiple-value-setq (val-data ,vidx)
                                        (deserialize-prim ,val-class ,vbuf ,vidx)))
                                    ((typep val-msg 'message-descriptor)
                                     (if (eq (proto-message-type val-msg) :group)
                                         (let ((tag2 (make-tag $wire-type-end-group 2)))
                                           `(multiple-value-setq (val-data ,vidx)
                                              ,(call-deserializer val-msg vbuf vidx nil tag2)))
                                         `(multiple-value-bind (payload-len payload-start)
                                              (decode-uint32 ,vbuf ,vidx)
                                            (setq ,vidx (+ payload-start payload-len)
                                                  val-data ,(call-deserializer val-msg vbuf
                                                                                'payload-start
                                                                                vidx)))))
                                    ((typep val-msg 'protobuf-enum)
                                     `(multiple-value-setq (val-data ,vidx)
                                        (deserialize-enum '(,@(protobuf-enum-values val-msg))
                                                          ,vbuf ,vidx)))
                                    ((typep val-msg 'protobuf-type-alias)
                                     (let ((class (proto-proto-type val-msg)))
                                       `(progn
                                          (multiple-value-setq (val-data ,vidx)
                                            (deserialize-prim ,class ,vbuf ,vidx))
                                          (setq ,val-data (funcall #',(proto-deserializer val-msg)
                                                                   ,temp)))))
                                    (t
                                     (undefined-field-type
                                      "While generating 'deserialize-object' for ~S,"
                                      message val-class field))))))))))

bkuehnert:
> What about:
Done

serialize.lisp
bkuehnert: This particular whitespace change should be fixed, as #63 was merged into this branch. However, there are similar whitespace changes as I fixed the other serialization functions as part of this branch.

text-format.lisp
bkuehnert: Changed in #74

text-format.lisp
bkuehnert: Changed in #74

define-proto.lisp
```diff
@@ -409,6 +410,47 @@ Parameters:
...
+                           (fintern "~A~A" msg-conc slot))))
+         (internal-slot-name (fintern "%~A" slot))
+         (mslot  (make-field-data
+                  :internal-slot-name internal-slot-name
+                  :external-slot-name slot
+                  :type 'hash-table
+                  :initform '(make-hash-table)
+                  :accessor reader))
+         (mfield (make-instance 'field-descriptor
+                  :name (slot-name->proto slot)
```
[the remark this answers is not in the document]
bkuehnert: This is a really strange one. Slot is currently bound to whatever gets passed in as the name from the protoc compiler, not (proto->slot-name name *package*). What's weird is that if you have a field called "map_field", protoc will lisp-ize it and give you "map-field" as the name. So, this function slot-name->proto is necessary as it's actually undoing what the protoc compiler does.

What's more confusing about this to me is the fact that the correspondence between "proto names" and "lisp names" is not one-to-one. What would happen if someone chose to make fields called MapField and map_field? Of course no one should ever do that, but the fact that it will cause an error should be documented.

protoc/field.cc
```diff
@@ -209,23 +209,35 @@ void GenerateField(io::Printer* printer, const FieldDescriptor* field) {
...
+  if(field->is_map()) {
+    vars["type"] = StrCat("(proto:map-of ",
+                          FieldLispType(field->message_type()->field(0)),
+                          " ",
+                          FieldLispType(field->message_type()->field(1)),
+                          ")");
+    printer->Print(vars,
+                    "\n(proto:define-map $name$\n"
+                    "  (:type $type$\n"
+                    "   :index $tag$))");
```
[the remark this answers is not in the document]
bkuehnert: Ended up going with (define-map foo int32 string :index n). The map-of part wasn't necessary.

model-classes.lisp
```diff
@@ -351,6 +359,13 @@ Parameters:
          (end2   (if (eql (char name2 0) #\() (- (length name2) 1) (length name2))))
     (string= name1 name2 :start1 start1 :end1 end1 :start2 start2 :end2 end2)))
 
+(defstruct map-descriptor
```
Slids, line 362: please put this near *maps* and find-maps

serialize.lisp
```diff
@@ -295,6 +296,28 @@ Parameters:
...
+                      (let ((ret-len (encode-uint32 tag buffer))
+                            (map-len 0))
+                        (with-placeholder (buffer)
+                          ; key types are always scalar, so serialize-prim works.
+                          (iincf map-len (serialize-prim k key-class
+                                                         (make-tag key-class 1) buffer))
+                          ; value types are arbitrary, non-map, non-repeated.
+                          (iincf map-len (emit-non-repeated-field v val-class 2 buffer))
+                          ;todo(benkuehnert): + or i+ here?
+                          (+ ret-len (+ map-len (backpatch map-len)))))))
```
Slids, line 317: Probably i+, try it and run a test, but don't spend to much time if it fails.

serialize.lisp
```diff
@@ -590,74 +613,262 @@ See field-descriptor for the distinction between index, offset, and bool-number.
         (if (not cell) ; skip this field
             (setq index (skip-element buffer index tag))
             ;; cell = (#<field> DATA . more) - "more" is the tail of the plist
+            ;; CELL now points to the cons where DATA should go.
```
Slids, line 616: Since you edited the function you need to add a docstring.

serialize.lisp
```diff
@@ -590,74 +613,262 @@ See field-descriptor for the distinction between index, offset, and bool-number.
...
+                                 val-class buffer index map-tag nil nil))))))
+                    (rplaca cell
+                            (progn
+                              (multiple-value-setq (data index)
+                                (deserialize-structure-object-field
+                                 type buffer index tag repeated-p lazy-p cell))
+                              data))))))))))
+
+
+(defun deserialize-structure-object-field
```
Slids, line 661: needs docstring

serialize.lisp
```diff
@@ -668,95 +879,17 @@ See field-descriptor for the distinction between index, offset, and bool-number.
...
-                   (let* ((class (proto-proto-type msg))
-                          (tag   (make-tag class (proto-index field))))
-                     `(let ((,vval ,reader))
-                        (when ,vval
-                          (let ((,vval (funcall #',(proto-serializer msg) ,vval)))
-                            (iincf ,size (serialize-prim ,vval ,class ,tag ,vbuf)))))))
-                  (t
-                   (undefined-field-type "While generating 'serialize-object' for ~S,"
-                                         msg class field))))))))
+            (let ((vector-p (vector-field-p field))
```
Slids, line 882: since you edited the function you need to add a docstring

serialize.lisp
```diff
@@ -806,136 +939,229 @@ Parameters:
...
   (let ((nslot nil)
         (rslot nil))
-    (flet ((ret (tag collector)
-             (return-from generate-field-deserializer (values (list tag) (list collector)
-                                                              nslot rslot)))
-           (ret-list (tags collectors)
-             (return-from generate-field-deserializer (values tags collectors
-                                                              nslot rslot)))
-           (call-deserializer (msg vbuf start end &optional (end-tag 0))
+    (let* ((class  (proto-class field))
```
Slids, line 942: again docstring

tests/map-test.lisp
```diff
@@ -0,0 +1,111 @@
...
+    (assert (= (slot-value result 'clunit::errors) 0))))
+
+;; Ideas for testing
+;; x Accessor function check
+;; - Serialize/deserialize pipe (check for raw bytes in middle)
+;; - Ensure there is a good error when we serialize a bad hashmap.
+;; - Optimized serialization
+;; - Print and ensure keys are sorted
+;; x Hashing to more complicated types (messages, nested messages)
+;; x ^ ensure we return right default value.
```
Slids, line 34: I don't understand what this means.
Add comments to your tests saying what they do
and if you want more tests make a todo

bkuehnert: This shouldn't be here at all. I'll make those changes

model-classes.lisp
bkuehnert: Right now, all of the "find-*" functions are together, and then the defstructs come after. So, this is currently consistent with the way the file is laid out. Do you think we should re-organize the file in a separate PR though?

define-proto.lisp
```diff
@@ -539,6 +592,64 @@ Arguments:
...
+          ; (or null [message-type]). This ensures that we type-check ,new-val against
+          ; the message type and disallow inserting nil into the map.
+          ,(if (listp val-type)
+               `(declare (type ,(third val-type) ,new-val))
+               `(declare (type ,val-type ,new-val)))
+          (setf (bit (,is-set-accessor ,obj) ,index) 1)
+          (setf (gethash ,new-key (,hidden-accessor-name ,obj)) ,new-val))
+
+        (declaim (inline ,public-remove-name))
+        (defun ,public-remove-name (,obj ,new-key)
```
cgay: I feel like the parameter ordering for the getter and remover should probably be the same as for gethash and remhash, and the setter should be in `(setf (getter k t) v)` form. Can do this in a follow-up PR. (Or not do it.)

model-classes.lisp
```diff
@@ -230,6 +230,14 @@ Parameters:
     (or (find-message type)
         (find-type-alias type))))
 
+(defvar *maps* (make-hash-table :test 'eq)
+  "Maps map names (symbols) to map-descriptor instances.")
+
+(declaim (inline find-map))
+(defun find-map (type)
```
cgay, line 237: Could you call this `find-map-descriptor`?

model-classes.lisp
Slids: Nope, nvm it's okay.

model-classes.lisp
```diff
@@ -351,6 +359,13 @@ Parameters:
          (end2   (if (eql (char name2 0) #\() (- (length name2) 1) (length name2))))
     (string= name1 name2 :start1 start1 :end1 end1 :start2 start2 :end2 end2)))
 
+(defstruct map-descriptor
+  (class     nil :type (or null symbol))
```
cgay, line 363: Do these slots really all need nullable types?  It looks like values are supplied for all of them when `make-map-descriptor` is called in `define-map` (but possibly some of those values are NIL?).

utilities.lisp
```diff
@@ -249,24 +249,35 @@
 Arguments:
   PROTO-TYPE: The symbol naming a protobuf message, group, etc.
   SLOT: The symbol naming a protobuf field.
-  FUNCTION-TYPE: The type of function name to retrieve: :has, :get, or :clear"
+  FUNCTION-TYPE: The type of function name to retrieve: :has, :get, or :clear.
+                 This can also be :map-get, :map-put, or :map-rem the for special map functions."
```
cgay, line 253: `the for` -> `for the`

define-proto.lisp
bkuehnert: I agree with the parameter ordering, will fix.

For the (setf (getter k t) v) form: So, getter in this case is whatever public-getter-name is bound to? I didn't realize you could do this with setf.

model-classes.lisp
bkuehnert: Sure, that's a more _descriptive_ name. Are we updating find-{message,enum,field} as well?

define-proto.lisp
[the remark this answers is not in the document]
bkuehnert: This is for a separate PR, correct?

model-classes.lisp
bkuehnert: No, these values should never be nil. Will update.

define-proto.lisp
```diff
@@ -411,6 +412,58 @@ Parameters:
...
+
+(defmacro define-map (type-name &key key-type val-type index)
+"Define a lisp type given the data for a protobuf map type.
+
+Parameters:
+  TYPE-NAME: Map type name.
+  KEY-TYPE: The lisp type of the map's keys.
+  VAL-TYPE: The lisp type of the map's values.
+  INDEX: Index of this map type in the field."
+  (check-type index integer)
```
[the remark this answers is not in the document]
bkuehnert: I'll add this to my list of things to do when stuck. Besides type checking, it seems extra useful in documentation where we differentiate between the index of a buffer and the index of a field.

bkuehnert:
All of the remaining comments seem to be out of scope for this PR. I'm going to leave them un-resolved and come back to them when stuck on other things.

However, there are two remaining things to be done with map types:
1. Type checking no longer works for adding to a map. The defsetf syntax is still strange to me, and I can't figure out how to add a declaration that would properly check types. My attempt is commented out on lines 620 & 621 of define-proto.lisp. Any input is much appreciated
2. Add text format support. This will be done in a separate PR. The basics are in this PR, although there isn't support for parsing yet.

define-proto.lisp
```diff
@@ -539,6 +591,67 @@ Arguments:
...
+
+      ; If the map's value type is a message, then the default value returned
+      ; should be nil. However, we do not want to allow the user to insert nil
+      ; into the map, so this binding only applies to the clear and get functions.
+      ,@(let ((val-type (if (find-message val-type)
+                            (list 'or 'null val-type)
+                            val-type))
+              (obj (gensym "OBJ"))
+              (new-key (gensym "NEW-KEY"))
+              (new-val (gensym "NEW-VAL")))
```
Slids, line 633: this is unused

define-proto.lisp
```diff
@@ -539,6 +591,67 @@ Arguments:
...
+      (declaim (inline (setf ,public-accessor-name)))
+      (defsetf ,public-accessor-name (new-key obj) (new-val)
+        `(progn
+           ; todo(benkuehnert): get this type checking working
+           ;(declare (type ,',key-type new-key))
+           ;(declare (type ,',val-type new-val))
+           (setf (bit (,',is-set-accessor ,obj) ,',index) 1)
+           (setf (gethash ,new-key (,',hidden-accessor-name ,obj)) ,new-val)))
+
+      ; If the map's value type is a message, then the default value returned
```
Slids, line 625: use two not one semicolons

define-proto.lisp
```diff
@@ -539,6 +591,67 @@ Arguments:
...
+         (key-type (map-descriptor-key-type (find-map-descriptor (proto-class field))))
+         (val-type (map-descriptor-val-type (find-map-descriptor (proto-class field))))
+         (val-default-form (get-default-form val-type $empty-default))
+         (is-set-accessor (fintern "~A-%%IS-SET" proto-type))
+         (index (proto-field-offset field)))
+
+    `(
+      (declaim (inline (setf ,public-accessor-name)))
+      (defsetf ,public-accessor-name (new-key obj) (new-val)
+        `(progn
```
Slids, line 618: why do you have this line?

define-proto.lisp
```diff
@@ -539,6 +591,67 @@ Arguments:
...
+           (setf (gethash ,new-key (,',hidden-accessor-name ,obj)) ,new-val)))
+
+      ; If the map's value type is a message, then the default value returned
+      ; should be nil. However, we do not want to allow the user to insert nil
+      ; into the map, so this binding only applies to the clear and get functions.
+      ,@(let ((val-type (if (find-message val-type)
+                            (list 'or 'null val-type)
+                            val-type))
+              (obj (gensym "OBJ"))
+              (new-key (gensym "NEW-KEY"))
```
Slids, line 632: why not just use with-gensyms like the other make-* functions do

define-proto.lisp
```diff
@@ -539,6 +591,67 @@ Arguments:
...
+         (val-default-form (get-default-form val-type $empty-default))
+         (is-set-accessor (fintern "~A-%%IS-SET" proto-type))
+         (index (proto-field-offset field)))
+
+    `(
+      (declaim (inline (setf ,public-accessor-name)))
+      (defsetf ,public-accessor-name (new-key obj) (new-val)
+        `(progn
+           ; todo(benkuehnert): get this type checking working
+           ;(declare (type ,',key-type new-key))
```
Slids, line 620: why do you have the quote?

Slids:
> All of the remaining comments seem to be out of scope for this PR. I'm going to leave them un-resolved and come back to them when stuck on other things.
> 
> However, there are two remaining things to be done with map types:
> 
> 1. Type checking no longer works for adding to a map. The defsetf syntax is still strange to me, and I can't figure out how to add a declaration that would properly check type. See lines 620 & 621 of define-proto.lisp. I've also tried (declare (type ,',key-type ,new-key)) yet that expands to (declare (type proto:int32 0)), if say key-type is proto:int32 and new-key happens to be 0, and that doesn't work. The defsetf macro binds new-key to the inputted value before anything gets executed, so I'm not sure how to splice in a type check. Any input/advice is much appreciated.
> 2. Add text format support. This will be done in a separate PR. The basics are in this PR, although there isn't support for parsing yet.

For 1 try:
(defun make-map-accessor-forms (proto-type public-slot-name slot-name field)
  "This creates forms that define map accessors which are type safe. Using these will
guarantee that the resulting map can be properly serialized, whereas if one modifies
the underlying map (which is accessed via the make-common-forms-for-structure-class
function) then there is no guarantee on the serialize function working properly.

Arguments:
  PROTO-TYPE: The Lisp type name of the proto message.
  PUBLIC-SLOT-NAME: Public slot name for the field (without the #\% prefix).
  SLOT-NAME: Slot name for the field (with the #\% prefix).
  FIELD: The class object field definition of the field."
  (let* ((public-accessor-name (proto-slot-function-name proto-type public-slot-name :map-get))
	 (public-remove-name (proto-slot-function-name proto-type public-slot-name :map-rem))
	 (hidden-accessor-name (fintern "~A-~A"  proto-type slot-name))
	 (key-type (map-descriptor-key-type (find-map-descriptor (proto-class field))))
	 (val-type (map-descriptor-val-type (find-map-descriptor (proto-class field))))
	 (val-default-form (get-default-form val-type $empty-default))
	 (is-set-accessor (fintern "~A-%%IS-SET" proto-type))
	 (index (proto-field-offset field)))

    (with-gensyms (obj new-val new-key)
      `(
	(declaim (inline (setf ,public-accessor-name)))
	(defsetf ,public-accessor-name (new-key obj) (,new-val)
	  ;; todo(benkuehnert): get this type checking working
	  (declare (type ,key-type new-key)
		   (type ,val-type new-val))
	  (setf (bit (',is-set-accessor ,obj) ',index) 1)
	  (setf (gethash ,new-key (',hidden-accessor-name ,obj)) ,new-val))

      ; If the map's value type is a message, then the default value returned
      ; should be nil. However, we do not want to allow the user to insert nil
      ; into the map, so this binding only applies to the clear and get functions.
      ,@(let ((val-type (if (find-message val-type)
			    (list 'or 'null val-type)
			    val-type))
	      ;;(new-val (gensym "NEW-VAL"))
	      )
	  `((declaim (inline ,public-accessor-name))
	    (defun ,public-accessor-name (,new-key ,obj)
	      (declare (type ,key-type ,new-key))
	      (the (values ,val-type t)
		   (multiple-value-bind (val flag)
		       (gethash ,new-key (,hidden-accessor-name ,obj))
		     (if flag
			 (values val flag)
			 (values ,val-default-form nil)))))

	    (declaim (inline ,public-remove-name))
	    (defun ,public-remove-name (,new-key ,obj)
	      (declare (type ,key-type ,new-key))
	      (remhash ,new-key (,hidden-accessor-name ,obj))
	      (if (= 0 (hash-table-count (,hidden-accessor-name ,obj)))
		  (setf (bit (,is-set-accessor ,obj) ,index) 0)))))

	(export '(,public-accessor-name
		  ,public-remove-name))))))

also, two semicolons not one.

define-proto.lisp
bkuehnert: My understanding is that the "long-form" defsetf describes how setf is macro-expanded. So, we should be returning code here. It's done like this in all of the examples I can find.

define-proto.lisp
bkuehnert: Because defsetf already binds the argument variables with gensym. http://www.lispworks.com/documentation/lw70/CLHS/Body/m_defset.htm

PR #73, opened 2020-06-30
Fix compiler warnings in serialization-test

bkuehnert:
Here we create an optimized serializer for color-wheel2, but don't
for color2 (which appears as a submessage). I'm not sure if you are
allowed to do this or not, but the compiler complained about an
undefined function (even though the tests passed before this change).
I'm also not sure why I got this warning now and not when I first
added the test.

This commit simply adds the call to make the serializer for color2,
which gets rid of the warning.


Slids:
Why doesn't this have cla?

bkuehnert:
> Why doesn't this have cla?

That's bizarre. I just sent in another PR and the CLA was found.

PR #83, opened 2020-07-08
Add documentation for map types

bkuehnert:


README.md
```diff
@@ -291,6 +291,40 @@ The type-url must be of the form base-url/qualified-name.
...
+### Map Types
+
+This library supports map types. Defining a proto MSG with map field FIELD
+will generate the following functions:
+
+```lisp
+(defun msg.field-gethash (key object))
+```
+Return the value mapped from KEY in OBJECT. If there is no value set, this
+function returns the default value of the map's value type.
```
cgay, line 303: It's not totally clear what `object` is here. I think naming it `msg` instead would be enough.  Applies to the rest of the docs as well.

README.md
```diff
@@ -291,6 +291,40 @@ The type-url must be of the form base-url/qualified-name.
...
+
+```lisp
+(defun (setf (msg.field-gethash) (value key object)))
+```
+Set the hash of KEY to VALUE in OBJECT. An example of using this function
```
cgay, line 308: "Set the value ..."

README.md
```diff
@@ -291,6 +291,40 @@ The type-url must be of the form base-url/qualified-name.
...
+would be:
+```lisp
+(setf (msg.field-gethash key object) value)
+```
+Which is in line with Common Lisp's `hash-table`.
```
cgay, line 313: lowercase "Which" or change it to "This"

README.md
```diff
@@ -291,6 +291,40 @@ The type-url must be of the form base-url/qualified-name.
...
+Remove the hash of KEY in OBJECT.
+
+These three functions are type checked, and interfacing with the map with
+these three functions alone will guarantee that (de)serialization functions
+as well as the `msg.has-field` function will work properly. However, if
+the user wants access to the underlying hash-table then they may use the
+generic accessor `(msg.field (object))` defined above (where `field` is a
+map field). Warning: this accessor is not type checked, so undefined behaviour
+may occur.
+
```
cgay, line 327: Line 322-326 can be just "The underlying hash table may be accessed directly via `(msg.field msg)`, but doing so may result in undefined behavior if the wrong types are used."

README.md
bkuehnert: I was struggling to find a clear/concise way to describe what object is here. How about calling it msg, and saying "... from KEY in message object MSG ..." ?

bkuehnert: On second thought, renaming object to msg doesn't seem right, as msg is already used in the expression. So, maybe keep it as object, but write it as "... from KEY in message object OBJECT..." perhaps?

README.md
bkuehnert: I want to be sure to mention that the has-* functions won't work either. If the user accesses the underlying hash table and adds to it via `(setf (gethash key (msg.field msg)) value)`, then we have no way of updating the is-set vector to reflect that the map is no longer non-empty. However, if they make a new hashtable `ht` and update the proto message with `(setf (msg.field msg) ht)` then it'll work properly. 

We could get around this problem entirely by asserting that hash tables are always set, although it may make serialization a little bit slower, as it would be checking if the hash-table has any entries rather than just checking if a bit is 1 or 0.

README.md
cgay: The repetition is pretty common for Lisp. If you have a message of type `Config` it wouldn't be unlikely to see `(config.user-metadata-gethash "id" config)`. Maybe using a less generic example would be better?  Looking at existing protobuf maps they tend to be used for buckets of open-ended. things like "metadata", "attributes", "labels"

cgay: Whatever you come up with just mark this as resolved because it'll be fine or we can fix it later.

README.md
cgay: My main beef was with the English. Let's keep the docs in the impersonal form where we don't use "we" or "you" or even "the user" if we can avoid it. Generally it can be implied, as in the way I rephrased it above.

I think you're right about making it clear about has-* functions. Perhaps we shouldn't even provide a way to access the underlying hash table? It can always be accessed via `slot-value` if someone's really determined to have their code break when the underlying implementation is changed. 💯

README.md
```diff
@@ -291,6 +291,40 @@ The type-url must be of the form base-url/qualified-name.
 Creates an any message protobuf message given a protobuf MESSAGE
 and a BASE-URL.
 
+### Map Types
+
+This library supports map types. Defining a proto MSG with map field FIELD
+will generate the following functions:
+
+```lisp
+(defun msg.field-gethash (key object))
```
Slids, line 300: Should we add the (defmethod field-gethash ((key key-type) (object proto-msg-type))) 
like we do for the regulat fields?
Definitely in a different pr

bkuehnert: Maybe. From the README, it seems like that accessor is deprecated. What would be the benefit of adding it?

Slids: Why do you say that (I've been specifically fighting to keep the defmethods)

Slids: https://github.com/qitab/cl-protobufs/pull/92

bkuehnert: I'm open to adding the defmethod, but is it necessary? Why do we need two functions that do identical things?

Slids: Those functions can get amazingly long with nested mesasges.
the 
foo.bar.bas.quux-gethash but foo bar and baz are longer.

bkuehnert: Makes sense, I think it's a good idea to add them then.

Slids:
Updated in https://github.com/qitab/cl-protobufs/pull/105

PR #98, opened 2020-07-16
Add oneof types

bkuehnert:
This PR adds oneof types (http://b/154711647).

There are still two things yet to be done. Both will be added in separate PRs. 
1. Add support for lazy fields inside of a oneof.
2. Update the printer

This PR also makes two nontrivial changes to non-oneof code:
1. The lambda function on line 983 of define.proto now maps over fields rather than slots. Currently, the `field-descriptor` class AND the `field-data` struct both do basically the same thing. The only reason you would ever want a `field-data` struct instead of a `field-descriptor` is when you define a slot on the object which doesn't constitute a field (i.e. the `%%BOOL-VALUES` and `%%IS-SET` vectors). So in that sense, the name `field-data` is quite bad. In this particular function, it makes no sense to iterate over the `field-data `structs, since everything here is assumed to be a field. This probably should've been a separate PR, but it was necessary to prevent code duplication in oneofs.
2. The `make-symbol` call in generate-field-deserializer (line 1050, serialize.lisp) are now `fintern` calls. Previously, it was OK to have this be a make-symbol, since this would only be called once per field. However, fields in the same oneof all share the same internal slot name and hence need to look at the same symbol. I didn't see clean way of ensuring that each oneof field shared the same `(make-symbol)` call, short of restructuring the function entirely, so I decided to use `fintern` across the board. 

There will be a lot of linting errors with this PR, I figured it would be easier to go through all of them at once rather than trying to hunt down places where the linter would complain.


api.lisp
```diff
@@ -78,6 +78,31 @@ only if the same fields have been explicitly set."
...
+            = (slot-value message-1 (oneof-descriptor-internal-name oneof))
+          for slot-value-2
+            = (slot-value message-2 (oneof-descriptor-internal-name oneof))
+          for lisp-type
+            = (when slot-value-1 (proto-class
+                                  (aref (oneof-descriptor-fields oneof)
+                                        (oneof-data-set-field slot-value-1))))
+          when (not (and slot-value-1 slot-value-2))
+            do (when (or slot-value-1 slot-value-2)
+                 (return-from proto-equal nil))
```
Slids, line 92: (return-from proto-equal) ?

bkuehnert: Is that a shorthand for `(return-from proto-equal nil)`? I actually want to return nil if control reaches this point.

Slids: Yea, you don't need to specifiy nil. 
I know we did earlier, but we shouldn't have...

tests/oneof-test.lisp
```diff
@@ -0,0 +1,119 @@
...
+  INTERACTIVE-P: Open debugger on assert failure."
+  (let ((result (run-suite 'oneof-tests :use-debugger interactive-p)))
+    (print result)
+    (assert (= (slot-value result 'clunit::failed) 0))
+    (assert (= (slot-value result 'clunit::errors) 0))))
+
+;; Test that normal accessors for fields in a oneof work.
+(deftest accessor-check (oneof-tests)
+  (let ((msg (make-oneof-proto)))
+    (assert-true (not (oneof-proto.has-intval msg)))
```
Slids, line 34: assert-false?
and below

api.lisp
bkuehnert: Is there any reason other than style that you would not write `nil`? If not, I vote for keeping it, as it better explains what the code is doing. If it was `(return-from proto-equal)` originally, then it would've taken me a bit longer to understand.

define-proto.lisp
```diff
@@ -592,6 +642,89 @@ Arguments:
...
+                            ,(proto-oneof-offset field))
+                        (oneof-data-value (,hidden-accessor-name ,obj))
+                        ,default-form))
+
+                  (declaim (inline (setf ,public-accessor-name)))
+                  (defun (setf ,public-accessor-name) (,new-value ,obj)
+                    (declare (type ,field-type ,new-value))
+                    (setf (bit (,is-set-accessor ,obj) ,index) 1)
+                    (setf (oneof-data-set-field (,hidden-accessor-name ,obj))
+                          ,(proto-oneof-offset field))
```
Slids, line 704: bind in the let instead of multiple calls here.

define-proto.lisp
```diff
@@ -930,6 +1092,17 @@ Arguments:
                 (push extra-field (proto-fields msg-desc)))
                ((define-extension)
                 (push model (proto-extensions msg-desc))))))
+          ((define-oneof)
+           (push `(:field-offset ,field-offset)
```
Slids, line 1096: o.O what what

Slids: Please don't...

define-proto.lisp
```diff
@@ -1324,6 +1499,17 @@ Arguments:
                 (appendf (proto-fields message) (list extra-field)))
                ((define-extension)
                 (appendf (proto-extensions message) (list model))))))
+          ((define-oneof)
+           (push `(:field-offset ,field-offset)
+                 (cdr (nthcdr 0 field)))
```
Slids, line 1504: again, please don't...

model-classes.lisp
```diff
@@ -328,6 +328,28 @@ Parameters:
...
+  (set-field nil :type (or null (unsigned-byte 32))))
+
+(defstruct oneof-descriptor
+  "The meta-object for a protobuf map"
+  ;; While oneofs do not have indices, each of its fields has a 'field-offset'
+  ;; which tracks the order that the fields are defined in the message. Since
+  ;; these fields share a single slot, they share the same field-offset. This
+  ;; slot holds this value.
+  (index nil :type (unsigned-byte 29))
+  (fields nil :type array)         ; A vector which stores the oneof's field descriptors.
```
Slids, line 345: typed array?
Does it matter though...

model-classes.lisp
```diff
@@ -375,6 +397,11 @@ Parameters:
            :accessor proto-fields
            :initarg :fields
            :initform ())
+   ;; A list of all oneof descriptors defined in this message.
+   (oneofs :type (list-of oneof-descriptor)
+           :accessor proto-oneofs
+           :initarg :oneofs
+           :initform ())
```
Slids, line 404: nil please.

serialize.lisp
```diff
@@ -948,6 +1010,26 @@ Parameters:
...
+    `(let* ((oneof-data (slot-value ,vobj ',(oneof-descriptor-internal-name oneof)))
+            (set-field  (oneof-data-set-field oneof-data))
+            (value      (oneof-data-value oneof-data)))
+       (ecase set-field
+         ,@(loop for field across fields
+                 collect
+                 (let ((class (proto-class field))
+                       (index (proto-index field))
+                       (offset (proto-oneof-offset field)))
+                   ;; BOUNDP is always true here, since if we get to this point
```
Slids, line 1024: Generally true but what if I overide the boundp function somehere?
We do such silly things internally...
Maybe your correct?

serialize.lisp
```diff
@@ -1234,18 +1339,20 @@ Parameters:
                       ;; Nonrepeating slots
                       (nslots collect-nslot)
                       ;; For tracking repeated slots that will need to be reversed
-                      (rslots collect-rslot))
+                      (rslots collect-rslot)
+                      ;; For tracking oneof slots
+                      (oneofslots collect-oneofslot))
```
Slids, line 1344: oneof-slots collect-oneof-slot

serialize.lisp
```diff
@@ -1281,6 +1391,8 @@ Parameters:
                     (type array-index ,vidx ,vlim))
            (let (,@(loop for slot in nslots
                          collect `(,slot ,missing-value))
+                 ,@(loop for oneofslot in oneofslots
```
Slids, line 1394: oneof-slot

api.lisp
Slids: Standard lisp, you wouldn't write
(defun ret-nil () nil)
(defun ret-nil (values nil))
Youd write 
(defun ret-nil ()) or
(defun ret-nil (values))

You never specify your returning nil, it's the default.
Carls nitted me on this several times...

Slids:
Also, there was some serialization changes, benchmarks?

model-classes.lisp
```diff
@@ -328,6 +328,28 @@ Parameters:
 (defmethod make-load-form ((m map-descriptor) &optional environment)
   (make-load-form-saving-slots m :environment environment))
 
+(defstruct oneof-data
+  (value nil)
+  ;; This slot stores data to tell which field is set in the oneof.
+  ;; It is either nil or a number. If it is nil, then nothing is set.
+  ;; if it is a number, say N, then the N-th field in the oneof is set.
+  (set-field nil :type (or null (unsigned-byte 32))))
```
Slids, line 336: please make signed-byte and -1 as unset
not have or null
and make a comment.

model-classes.lisp
```diff
@@ -328,6 +328,28 @@ Parameters:
 (defmethod make-load-form ((m map-descriptor) &optional environment)
   (make-load-form-saving-slots m :environment environment))
 
+(defstruct oneof-data
+  (value nil)
```
Slids, line 332: comment that trying to type value has no size benefit
and has a huge code complexity increase

Slids:
Also, text-format in and out

define-proto.lisp
```diff
@@ -592,6 +642,89 @@ Arguments:
...
+         (hidden-accessor-name (fintern "~A-~A" proto-type hidden-slot-name))
+         (index (oneof-descriptor-index oneof))
+         (case-function-name (proto-slot-function-name proto-type public-slot-name :case))
+         (has-function-name (proto-slot-function-name proto-type public-slot-name :has))
+         (clear-function-name (proto-slot-function-name proto-type public-slot-name :clear)))
+    (with-gensyms (obj)
+      `(
+        (declaim (inline ,case-function-name))
+        (defun ,case-function-name (,obj)
+          (case (oneof-data-set-field (,hidden-accessor-name ,obj))
```
bkuehnert, line 658: ecase

api.lisp
cgay: uh, I am not sure what I bugged you about specifically, but in the case where you are explicitly returning from a function, saying what value you're returning seems like a good thing. I would use `(return from proto-equal nil)`. Not specifying a value is only okay if the function doesn't return any values, in my opinion.

cgay: Careful... returning `(values)` is distinct from returning an empty body.  
```
CL-USER> (defun foo ())                                                                                                                                                                                                                                                                  
FOO                                                                                                                                                                                                                                                                                      
CL-USER> (multiple-value-list (foo))                                                                                                                                                                                                                                                     
(NIL)                                                                                                                                                                                                                                                                                    
CL-USER> (defun foo () (values))                                                                                                                                                                                                                                                         
FOO                                                                                                                                                                                                                                                                                      
CL-USER> (multiple-value-list (foo))                                                                                                                                                                                                                                                     
NIL
```

serialize.lisp
```diff
@@ -117,8 +117,19 @@
...
-    (dolist (field (proto-fields msg-desc) size)
-      (iincf size (emit-field object field buffer)))))
+    (dolist (field (proto-fields msg-desc))
+      (iincf size (emit-field object field buffer)))
+    (dolist (oneof (proto-oneofs msg-desc) size)
+      (let* ((fields    (oneof-descriptor-fields oneof))
+             (data      (slot-value object (oneof-descriptor-internal-name oneof)))
+             (set-field (oneof-data-set-field data))
+             (value     (oneof-data-value data)))
+        (unless (eq set-field :unset)
```
cgay, line 127: Should this be `(when set-field` ?  It looks like `oneof-data-set-field` should return nil or an integer.  If I'm right then there probably isn't a test in which the oneof field isn't set.

bkuehnert: You're right. I've fixed this and added a test for unset oneof fields. Thank you

serialize.lisp
```diff
@@ -117,8 +117,19 @@
 (defmethod serialize-object (object (msg-desc message-descriptor) buffer)
   (declare (buffer buffer))
   (let ((size 0))
-    (dolist (field (proto-fields msg-desc) size)
-      (iincf size (emit-field object field buffer)))))
+    (dolist (field (proto-fields msg-desc))
+      (iincf size (emit-field object field buffer)))
+    (dolist (oneof (proto-oneofs msg-desc) size)
```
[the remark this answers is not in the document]
bkuehnert: I did consider that. IMO, oneofs are a separate appendage onto a message. They aren't fields, they are a container for fields. 

The only attribute that oneof-descriptor shares with field-descriptor is the fact that they both have a name, and both can be (de)serialized independent of the rest of the message. Index doesn't really apply here, since a oneof's index isn't the same as a field's index (field number?). 

And one final reason is separating oneofs from fields let me use a defstruct rather than a defclass.

define-proto.lisp
```diff
@@ -948,6 +1121,7 @@ Arguments:
                      ;; TODO(cgay): this should probably refer to the external field name but I'll
                      ;; wait since I've no idea if that slot is bound at this point.
                      (proto-internal-field-name field) (proto-class msg-desc))
+             ;; todo(benkuehnert: What does this do?
```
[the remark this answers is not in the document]
bkuehnert: I'm commenting here specifically on the (setq index idx) call. I still have no idea why it's there. It seems like index is declared locally at the top of the macro, then gets set a few different times, and then gets forgotten about. I haven't checked to make sure that it's not being secretly used in some other macroexpansion, though.

define-proto.lisp
```diff
@@ -930,6 +1092,17 @@ Arguments:
                 (push extra-field (proto-fields msg-desc)))
                ((define-extension)
                 (push model (proto-extensions msg-desc))))))
+          ((define-oneof)
+           (push `(:field-offset ,field-offset)
+                 (cdr (nthcdr 0 field)))
+           (destructuring-bind (&optional progn oneof-desc)
+               (macroexpand-1 field env)
+             (assert (eq progn 'progn) ()
+                     "The macroexpansion for ~S failed" field)
```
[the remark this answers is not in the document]
bkuehnert: Is that a more useful error message though? Is there a reason why the first value returned by the macro is progn, other than just indicating that the macro-expansion succeeded?

define-proto.lisp
```diff
@@ -791,7 +925,9 @@ Arguments:
                        (find-if #'(lambda (el)
                                     (eq (field-data-internal-slot-name el) '%%is-set))
                                 slots)))
-         (additional-slots '(%%is-set)))
+         (additional-slots '(%%is-set))
+         (oneof-fields (loop for oneof in oneofs
+                             append (coerce (oneof-descriptor-fields oneof) 'list))))
```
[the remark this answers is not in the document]
bkuehnert: The reason it's a vector is that I would like fast access to the field descriptor from the set-field (previously it was a hash-table, and I changed it to a vector to go faster). Is there a way to cheat lisp into giving me fast random access into a list?

model-classes.lisp
```diff
@@ -328,6 +328,28 @@ Parameters:
...
+  "The meta-object for a protobuf map"
+  ;; While oneofs do not have indices, each of its fields has a 'field-offset'
+  ;; which tracks the order that the fields are defined in the message. Since
+  ;; these fields share a single slot, they share the same field-offset. This
+  ;; slot holds this value.
+  (index nil :type (unsigned-byte 29))
+  (fields nil :type array)         ; A vector which stores the oneof's field descriptors.
+  (internal-name nil :type symbol) ; This is the external name, but with '%' prepended.
+  (external-name nil :type symbol) ; A symbol whose name is the name of the oneof.
+  (reader nil :type symbol))       ; A reader used to access the oneof-data struct.
```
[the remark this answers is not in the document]
bkuehnert: I like that better. Done.

model-classes.lisp
```diff
@@ -328,6 +328,28 @@ Parameters:
 (defmethod make-load-form ((m map-descriptor) &optional environment)
   (make-load-form-saving-slots m :environment environment))
 
+(defstruct oneof-data
```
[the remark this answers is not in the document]
bkuehnert: I'm... not totally sure why I didn't call it `oneof` to begin with. Switching to that now.

api.lisp
```diff
@@ -78,6 +78,31 @@ only if the same fields have been explicitly set."
...
+          when (not (equal (oneof-data-set-field slot-value-1)
+                           (oneof-data-set-field slot-value-2)))
+            do (return-from proto-equal nil)
+          when (or (scalarp lisp-type) (find-enum lisp-type))
```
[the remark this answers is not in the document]
bkuehnert: Wouldn't that be a lie? Enums are not scalars. Iinstead, maybe `scalar-field-equal` should be renamed to something that includes enums. Is there anything less awkward than `scalar-or-enum-field-equal` ?

define-proto.lisp
```diff
@@ -464,6 +463,57 @@ Parameters:
...
 
+(defmacro define-oneof ((&key field-offset) oneof-slot &body fields)
+  "Define a protobuf oneof. This macro creates the representation for
+the oneof, as well as the representation/defining forms for its fields.
+Warning: This macro assumes the variables 'conc-name', 'field-offset',
+and 'index' are bound when macroexpanding.
+
+Parameters:
+  NAME: The name of the oneof.
+  FIELDS: Field as output by protoc."
```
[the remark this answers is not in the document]
bkuehnert: Good point about the NAME. I've updated that.

As for why (:field-offset n) precedes the name: That's from a hack that will be removed soon

define-proto.lisp
```diff
@@ -464,6 +463,57 @@ Parameters:
...
+                                                         *protobuf* (or name
+                                                                        (slot-name->proto slot)))
+                                        :label :optional
+                                        :index index
+                                        :internal-field-name internal-name
+                                        :external-field-name slot
+                                        :field-offset field-offset
+                                        :oneof-offset oneof-offset
+                                        :default default
+                                        :lazy (and lazy t) ; todo(benkuehnert): why AND with t?
```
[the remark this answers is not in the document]
bkuehnert: Ahhh I see.

serialize.lisp
bkuehnert: The `BOUNDP` that I'm referencing is not a function here. I'm referring to the fact that when I call generate-non-repeated-field-serializer, I have `t` as the `BOUNDP` argument. This is due to the fact that if we get to this point (i.e. there is a non-nil value in the `set-field` slot, then it must be the case that there's a valid value bound to `value`.

model-classes.lisp
bkuehnert: Initform should be nil? The other slots which hold a list all have `:initform ()`. Are you suggesting to change those as well?

model-classes.lisp
bkuehnert: I added a type... No idea how much this matters, but I guess it can't hurt.

Slids:
Warning:
CheckLint found errors.
Warning: messages should be named using CamelCase. If this is a legacy file, consider adding "// LINT: LEGACY_NAMES" to the top of the file. [naming_conventions]
	//depot/google3/third_party/lisp/cl_protobufs/tests/oneof-proto.proto:21:10
No docstring for struct ONEOF. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/model-classes.lisp:331
No docstring for function GENERATE-ONEOF-SERIALIZER. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/serialize.lisp:1013
No mention of parameter FIELD-OFFSET in docstring for macro DEFINE-ONEOF. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/define-proto.lisp:466

model-classes.lisp
bkuehnert: I've pushed a commit which does this, but I'm not sure if this is worth it. There are no space savings on the object, and it's hard to say if there is any meaningful speed increase. The only place where there would be a speedup is in the `msg.oneof-case` function, but as you said: it's not used in (de)serialization so we don't need to squeeze every last drop of speed out of it.

The reason I say it's not worth it is because it makes the code more complicated and harder to read. Previously, if we want to predicate on whether or not a oneof is set, we would write `(if set-field ...)`. Now, you have to write `(if (not (equal set-field -1)) ...)` or some other function which checks if a number is non-negative. Either way, it is more confusing to read.

api.lisp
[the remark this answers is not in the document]
Slids: I disagree with carl, why should scalerp return t for enums?

model-classes.lisp
Slids: Sounds good.

Slids: To much work, nvm.

define-proto.lisp
```diff
@@ -592,6 +642,89 @@ Arguments:
...
+      `(
+        (declaim (inline ,case-function-name))
+        (defun ,case-function-name (,obj)
+          (case (oneof-data-set-field (,hidden-accessor-name ,obj))
+            ,@(loop for field across (oneof-descriptor-fields oneof)
+                    collect
+                    (let ((offset (proto-oneof-offset field))
+                          (keyword (keywordify (proto-external-field-name field))))
+                      `((,offset) ,keyword)))
+            ((nil) :%unset)))
```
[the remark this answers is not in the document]
bkuehnert: This has already been updated. Github must be showing an old version.

define-proto.lisp
```diff
@@ -464,6 +463,57 @@ Parameters:
...
+         (desc (make-oneof-descriptor
+                :internal-name internal-name
+                :external-name oneof-slot
+                :fields (make-array (length fields)
+                                    :element-type 'field-descriptor)
+                :index field-offset))
+         (oneof-offset 0))
+    (loop for field in fields
+          do (destructuring-bind (slot &key type typename name (default nil default-p)
+                                         lazy index documentation &allow-other-keys)
```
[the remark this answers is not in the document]
bkuehnert: Emacs is telling me that this is the correct indentation. Are you saying that `lazy` should be aligned with `type`?

model-classes.lisp
Slids: SGTM

Slids:
Current lint:
Warning: messages should be named using CamelCase. If this is a legacy file, consider adding "// LINT: LEGACY_NAMES" to the top of the file. [naming_conventions]
	//depot/google3/third_party/lisp/cl_protobufs/tests/oneof-proto.proto:21:10
No docstring for struct ONEOF. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/model-classes.lisp:327
No docstring for function GENERATE-ONEOF-SERIALIZER. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/serialize.lisp:1014

serialize.lisp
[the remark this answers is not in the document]
bkuehnert: Can I resolve this?

model-classes.lisp
bkuehnert: (I'm unresolving this because I actually need to revert the commit first)

define-proto.lisp
```diff
@@ -592,6 +645,108 @@ Arguments:
...
+                    (and (= (bit (,is-set-accessor ,obj) ,index) 1)
+                         (eq (oneof-set-field (,hidden-accessor-name ,obj))
+                             ,oneof-offset)))
+
+                  (declaim (inline ,clear-function-name))
+                  (defun ,clear-function-name (,obj)
+                    (when (,has-function-name ,obj)
+                      (setf (bit (,is-set-accessor ,obj) ,index) 0)
+                      (setf (oneof-value (,hidden-accessor-name ,obj)) nil)
+                      (setf (oneof-set-field (,hidden-accessor-name ,obj)) -1)))
```
Slids, line 738: the has-function-name and clear-function-name do a LOT more work then they should.
Just Check if the field set on the oneof structure is you.
If it is, your set, if not, your not.
Should make things a bit more simple and save space.

define-proto.lisp
```diff
@@ -1351,10 +1567,11 @@ Arguments:
        (make-field-data
           :internal-slot-name 'proto-impl::%%is-set
           :external-slot-name 'proto-impl::%%is-set
-          :type `(bit-vector ,(length fields))
+          :type `(bit-vector ,(+ (length fields) (length oneofs)))
```
Slids, line 1570: I don't see why this is necessary. 
The oneof struct contains all of the information you
need to know to know if a field is set.
Here you're adding possibly many bits...

define-proto.lisp
```diff
@@ -592,6 +645,108 @@ Arguments:
...
+            ,@(loop for field across (oneof-descriptor-fields oneof)
+                    collect
+                    (let ((offset (proto-oneof-offset field))
+                          (keyword (keywordify (proto-external-field-name field))))
+                      `((,offset) ,keyword)))
+            ((-1) :%unset)))
+
+        (declaim (inline ,has-function-name))
+        (defun ,has-function-name (,obj)
+          (= (bit (,is-set-accessor ,obj) ,index) 1))
```
Slids, line 680: (not (= (oneof-set-field (,hidden-accessor-name ,obj)) -1))

define-proto.lisp
```diff
@@ -592,6 +645,108 @@ Arguments:
...
+
+        (declaim (inline ,clear-function-name))
+        (defun ,clear-function-name (,obj)
+          (setf (oneof-value (,hidden-accessor-name ,obj)) nil)
+          (setf (bit (,is-set-accessor ,obj) ,index) 0)
```
Slids, line 685: not needed

define-proto.lisp
```diff
@@ -983,10 +1185,11 @@ Arguments:
        (make-field-data
         :internal-slot-name 'proto-impl::%%is-set
         :external-slot-name 'proto-impl::%%is-set
-        :type `(bit-vector ,(length fields))
+        :type `(bit-vector ,(+ (length fields) (length oneofs)))
```
Slids, line 1188: shouldn't need  (length oneofs)

define-proto.lisp
```diff
@@ -983,10 +1185,11 @@ Arguments:
...
         :initarg :%%is-set
-        :initform `(make-array ,(length fields) :element-type 'bit
-                                                :initial-element 0)))
+        :initform `(make-array ,(+ (length fields) (length oneofs))
```
Slids, line 1190: shouldn't need (length oneofs)

define-proto.lisp
Slids, line 1570: (length oneofs)

define-proto.lisp
```diff
@@ -1351,10 +1567,11 @@ Arguments:
        (make-field-data
           :internal-slot-name 'proto-impl::%%is-set
           :external-slot-name 'proto-impl::%%is-set
-          :type `(bit-vector ,(length fields))
+          :type `(bit-vector ,(+ (length fields) (length oneofs)))
           :initarg :%%is-set
-          :initform `(make-array ,(length fields) :element-type 'bit
-                                                  :initial-element 0)))
+          :initform `(make-array ,(+ (length fields) (length oneofs))
```
Slids, line 1572: (length oneofs)

tests/oneof-test.lisp
```diff
@@ -0,0 +1,133 @@
...
+
+(defun run (&optional interactive-p)
+  "Run all tests in the test suite.
+Parameters:
+  INTERACTIVE-P: Open debugger on assert failure."
+  (let ((result (run-suite 'oneof-tests :use-debugger interactive-p)))
+    (print result)
+    (assert (= (slot-value result 'clunit::failed) 0))
+    (assert (= (slot-value result 'clunit::errors) 0))))
+
```
Slids, line 26: Do you have a test where you set one field of a oneof 
and then set another field?
This should be fine but I don't see a test for it.

bkuehnert: Yes, that's what `oneof-accessor-check` is supposed to do (among other things).

define-proto.lisp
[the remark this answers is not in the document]
Slids: Going to resolve this as 6 days with no response.

define-proto.lisp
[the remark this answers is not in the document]
Slids: No

define-proto.lisp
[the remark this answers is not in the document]
Slids: Yes, because with progn it is valid lisp code.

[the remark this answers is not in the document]
cgay: Let's put it this way: there are about 6 places where we say generically "the macroexpansion failed" and trying to figure out macroexpansion backtraces is No Fun At All.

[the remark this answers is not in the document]
bkuehnert: I'm not sure how this can be improved, and at the moment I'm not totally sure what conditions would cause this assertion to be false. Aside from a runtime error in macro-expansion, is there any other way for the first value returned to **not** be `progn`? And if so, how can we get a better error message than returning the macro and saying "this failed"?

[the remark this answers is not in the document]
Slids: At least say:
 "The macroexpansion for ~S failed in DEFINE-MESSAGE"
will be a little bit better.

PR #104, opened 2020-07-22
Simplify text format code

bkuehnert:
This PR refactors text-format.lisp in two ways:
1. Breaks the `do-field` flet into two separate functions `print-repeated-field` and `print-non-repeated-field`. This code was very similar to the `serialize-object` function, so the refactor is similar. This will make it a lot easier to support maps/oneofs.
2. Gets rid of code duplication in `parse-text-format`. This function had the same block of code twice (one for repeated fields, one for non repeated fields). Now, that block of code is called `parse-field`. This will also be important for parsing oneof/map fields.

This PR also expands the test in `tests/text-format-tests.lisp`, but I'm not sure how important that is (since the same sort of thing is tested in the serialization tests).


bkuehnert:
I'm not totally sure why CI passed for 27af4db but not b51bad1. For the record, they both passed locally.

text-format.lisp
```diff
@@ -240,125 +271,84 @@ Parameters:
...
-                                         (setf (proto-slot-value object slot)
-                                               (funcall (proto-deserializer msg) val))))))
-                                  (t
-                                   (undefined-field-type "While parsing ~S from text format,"
-                                                         msg-desc type field)))))))))))
-    (declare (dynamic-extent #'deserialize))
-    (deserialize (proto-class msg-desc))))
+  (let ((object (make-instance (or (proto-alias-for msg-desc)
+                                   (proto-class msg-desc))))
+        (rslots ()))
```
cgay, line 283: Could you add a comment here, or rename this to `repeated-slots`? 
`; repeated slot names, held onto so the values can be nreversed.`

text-format.lisp
```diff
@@ -240,125 +271,84 @@ Parameters:
...
+        (if (null field)
+            (skip-field stream)
+            (multiple-value-bind (val error-p)
+                (parse-field type :stream stream)
+              (cond
+                (error-p
+                 (undefined-field-type "While parsing ~S from text format,"
+                                       msg-desc type field))
+                ((eq (proto-label field) :repeated)
+                 (when slot
```
cgay, line 308: Under what conditions can `slot` be nil here? Also below.

text-format.lisp
```diff
@@ -240,125 +271,84 @@ Parameters:
...
+                   (pushnew slot rslots)
+                   (push val (proto-slot-value object slot))))
+                (t
+                 (when slot
+                   (setf (proto-slot-value object slot) val))))))))))
+
+(defun parse-field (type &key (stream *standard-output*))
```
cgay, line 315: Was this meant to be `*standard-input*`?

text-format.lisp
```diff
@@ -240,125 +271,84 @@ Parameters:
...
+  "Parse data of type TYPE from STREAM. This function returns
+the object parsed. If the parsing fails, the function will
+return T as a second value."
+  (let ((msg (or (find-message type)
```
cgay, line 319: We should be using `desc` to name something that holds a descriptor that may not be a message-descriptor.

text-format.lisp
bkuehnert: Yes...

text-format.lisp
bkuehnert: `slot` will be nil if we fail to find a field with `(find-field msg-desc name)`. So, this exists for backwards compatibility when adding new fields. I added a comment to explain this.

text-format.lisp
```diff
@@ -26,140 +32,165 @@ Parameters:
...
+              'message-descriptor)
+       (when value
+         (let ((indent (+ indent 2)))
+           (print-text-format value :indent indent
+                                    :stream stream
+                                    :name (proto-name field)
+                                    :print-name print-name
+                                    :pretty-print pretty-print))))
+      ((typep msg 'enum-descriptor)
+       (when (and value (not (eql value (proto-default field))))
```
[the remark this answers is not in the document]
bkuehnert: Never.. Since we check the boundedness of value at the top of the function, this isn't necessary.

text-format.lisp
```diff
@@ -26,140 +33,163 @@ Parameters:
...
-(defun print-scalar (val type field stream indent)
+Parameters:
+  VALUE: The value in the field to print.
+  FIELD: The field-descriptor of the field.
+  INDENT: If supplied, indent the text by INDENT spaces.
+  STREAM: The stream to output to.
+  PRINT-NAME: Whether or not to print the name of the field.
+  PRETTY-PRINT: When true, print newlines and indentation."
+  (unless value
+    nil) ; If value is nil, there is nothing to do.
```
cgay, line 134: This won't do an early return; you'll need `(return-from print-non-repeated-field nil)`.  Same in `print-repeated-field`.

text-format.lisp
```diff
@@ -26,140 +33,163 @@ Parameters:
...
+
+Parameters:
+  VALUES: The list or vector of values in the field to print.
+  FIELD: The field-descriptor of the field.
+  INDENT: If supplied, indent the text by INDENT spaces.
+  STREAM: The stream to output to.
+  PRINT-NAME: Whether or not to print the name of the field.
+  PRETTY-PRINT: When true, print newlines and indentation."
+  (unless values
+    nil) ; If values is NIL, then there is nothing to do.
```
Slids, line 90: (return-from print-repeated-field nil)

Slids: also, why the newline?

bkuehnert: Which newline are you referring to?

text-format.lisp
bkuehnert: That's embarassing, not sure what I was thinking

text-format.lisp
Slids: (unless values
    nil) ; If values is NIL, then there is nothing to do.
The newline between values and nil should be removed.
But the code was changed,

PR #107, opened 2020-07-24
Remove FIELD arguments from print-enum, print-repeated-field, and print-non-repeated-field

bkuehnert:
This argument isn't at all necessary, since the only information that it's used for is:
1. The name of the field
2.  Printing the field-descriptor when there is a type error.

This makes the function's errors slightly less informative, but it also makes the functions more generic.  

I could also add FIELD as an optional parameter to allow for better errors when it's supplied, but I'm not sure how much that's worth.


text-format.lisp
```diff
@@ -102,42 +105,43 @@ Parameters:
...
          (doseq (v values)
                 (let ((v (funcall (proto-serializer desc) v)))
-                  (print-scalar v type (proto-name field) stream
+                  (print-scalar v type name stream
                                 (and pretty-print indent))))))
       (t
-       (undefined-field-type "While printing ~S to text format,"
-                             values type field)))))
+       (undefined-type "While printing ~S to text format,"
+                             values type)))))
```
cgay, line 123: indentation. probably just use one line for this call.

text-format.lisp
```diff
@@ -180,8 +182,8 @@ Parameters:
            (format stream "~&~VT}" (+ indent 2)))))
 
       (t
-       (undefined-field-type "While printing ~S to text format,"
-                             value type field)))))
+       (undefined-field "While printing ~S to text format,"
+                             value type)))))
```
cgay, line 186: same here

text-format.lisp
```diff
@@ -243,8 +245,9 @@ Parameters:
             do not write a newline."
   (when val
     (if indent
-      (format stream "~&~V,0T~A: " (+ indent 2) (proto-name field))
-      (format stream "~A: " (proto-name field)))
+      (format stream "~&~V,0T" (+ indent 2)))
+    (if name
```
cgay, line 249: it's somewhat arbitrary, but the style is to use `when` if there is no `else` clause.

conditions.lisp
```diff
@@ -44,6 +44,14 @@
     :type-name (prin1-to-string type)
     :field field))
 
+;; This is used when the field-descriptor is not accessible.
```
cgay, line 47: This leaves me wondering why the field descriptor would be inaccessible. Why not just make it so this case is impossible. There are two possibilities:

1. We control all the code and we assume we don't have this bug anywhere. There's no point defensively coding for bugs we shouldn't have.  (This is why I object to code that has `(and type ...)` when `type` should never be nil.

2. We don't control all the code (i.e., for some reason (?) `print-[non-]repeated-field` is exported) in which case we declare the type of the `type` argument to be something that prevents falling off the end of the `cond`.  Probably `descriptor`?

Maybe instead of `cond` we should be using `etypecase`?  But that's a bigger change because we don't have types that partition the entire space and because one of the `cond` tests has side-effects.  I'd be satisfied with a comment that explains why we need this special kind of error instead of just `(error "you found a bug in our code...please report it [here]")`.  The latter makes it much clearer to the reader of the code that this is an abnormal state to be in.

cgay: My guess is that the error classes in the code are way over-engineered... :)

conditions.lisp
```diff
@@ -44,6 +44,14 @@
...
+(defun undefined-type (format-control object type)
```
cgay, line 48: A better parameter order would be `(type format-control &rest format-args)` so that it isn't necessary to assume there's exactly one format arg in `format-control`.  Also because this is named `undefined-TYPE` it seems natural for `type` to be the first arg.

conditions.lisp
bkuehnert: I believe that the error exists for people who use decide to add custom types via the lisp_type extension. As long as that option exists, I don't think we can rule out this case ever happening.
 
We also can't type the `type` argument for the same reason we can't replace `cond` with `etypecase`. We don't have a type that is the union of keywords and descriptor classes. 

Would a comment at the site of the exception saying "This case only happens when the user specifies a custom type and does not support it" suffice?

cgay: Yup, that would do it.  Or perhaps better at the call site where the `cond` falls through. (Both places.)

We could define a type for `scalarp`, but I'm not saying that we should. :)

bkuehnert: Right, that's what I meant. A comment wherever we throw the exception. 

I think it would be a good idea to make a type for scalar values, but obviously that's for another time

cgay:
There are some lint errors:
```
CheckLint found errors.
No docstring for function UNDEFINED-TYPE. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/conditions.lisp:48
Lines should not end in whitespace. (WHITESPACE-AT-EOL)
	//depot/google3/third_party/lisp/cl_protobufs/text-format.lisp:133
Lines should not end in whitespace. (WHITESPACE-AT-EOL)
	//depot/google3/third_party/lisp/cl_protobufs/text-format.lisp:195
```

PR #108, opened 2020-07-24
Oneof text format support

bkuehnert:
This PR changes print-text-format so that it knows what to do with oneof fields.

No changes are necessary to parse-text-format. It Just Works which is nice.


tests/oneof-test.lisp
```diff
@@ -131,3 +131,34 @@ Parameters:
                 (assert-true (eq (oneof-proto.my-oneof-case msg2) 'strval))
                 (assert-true (string= (oneof-proto.strval msg2) "hi"))
                 (assert-false (oneof-proto.has-intval msg2))))))
+
+(deftest text-format-test (oneof-tests)
```
Slids, line 135: This is good but it makes the assumption we print correctly.
Could you make one that is a known valid string to parse?

bkuehnert: Done. I added it to the text-format-test.lisp file, since that seemed like a reasonable place to put it.

tests/oneof-test.lisp
```diff
@@ -131,3 +131,34 @@ Parameters:
...
+         (test2 (make-nested-oneof :outside 2 :nested test1))
+         (intlist (make-oneof-test.int-list :ints (list 1 2 3 4 5)))
+         (test3 (make-oneof-test :list-of-ints intlist))
+         (test4 (make-oneof-proto :outside 1)))
+
+    (let* ((text (with-output-to-string (s)
+                   (print-text-format test1 :stream s)))
+           (parsed (with-input-from-string (s text)
+                     (parse-text-format 'oneof-proto :stream s))))
+      (assert-true (proto:proto-equal test1 parsed)))
```
cgay, line 146: How about using
```lisp
(flet ((round-trip (proto) ...))
   (assert-true (proto:proto-equal test1 (round-trip test1)))
   (assert-true (proto:proto-equal test2 (round-trip test2)))
   ...
```

bkuehnert: Sure, will do

PR #109, opened 2020-07-24
Add documentation for oneofs

bkuehnert:


README.md
```diff
@@ -351,8 +351,29 @@ hash table may be accessed directly via `(dictionary.map-field dict)`, but doing
 so may result in undefined behavior.
 
 ### Oneof
+This section uses the following protobuf message as an example:
+```protocol-buffer
```
cgay, line 355: Internally, at least, all triple-backquotes need a blank line above them if you want them to be rendered with newlines instead of inline in the paragraph.

README.md
```diff
@@ -351,8 +351,29 @@ hash table may be accessed directly via `(dictionary.map-field dict)`, but doing
...
 
-TODO
+```lisp
+(setf (person.age bob) 5)
+```
+will set the `age` field of a `Person` object `bob` to `5`. Defining a oneof also creates
+two special functions:
+`(person.age-oneof-case bob)`
+This will return the lisp symbol corresponding to the field which is currently set. So, if
+we set `age` to `5`, then this will return `age`. If no field is set, this function will
```
cgay, line 373: I think for clarity this needs to say "will return the symbol `AGE`".

bkuehnert: True. Done

Slids:
Carl, do you want to manually enter this one?
Ready for Google won't work.

Slids:
Manually pulled in
https://github.com/qitab/cl-protobufs/pull/117

PR #115, opened 2020-07-27
Remove some repeated code from serializers

bkuehnert:
Since the method serialize-object does exactly what these two
chunks of code does, just call that.

### DIFF
--- a/serialize.lisp
+++ b/serialize.lisp
@@ -65,18 +65,6 @@
 
 ;;; Serialization
 
-;; Allow clients to add their own methods
-;; This is how we address the problem of cycles, e.g. -- if you have an object
-;; that may contain cycles, serialize the cyclic object using a "handle"
-;; TODO(jgodbout): Delete the geenric serialize-object
-;; and remove the type option for the serialize defuns.
-(defgeneric serialize-object (object type buffer)
-  (:documentation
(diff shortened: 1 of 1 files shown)


Slids:
You can easily fix the defmethod call by making it a regular function.
I know we say:

;; This is how we address the problem of cycles, e.g. -- if you have an object
;; that may contain cycles, serialize the cyclic object using a "handle"

but I don't think protocol buffers can have cycles so this is a the tail
wagging the pervorbial dog.

Sadly I'll have to do some internal updates, but thats the price of fixing stupid code...

serialize.lisp
```diff
@@ -113,8 +101,10 @@
 
 ;; Serialize the object using the given protobuf type
 
-;; The default method uses metadata from the message descriptor.
-(defmethod serialize-object (object (msg-desc message-descriptor) buffer)
+;; The default function uses metadata from the message descriptor.
+(defun serialize-object (object msg-desc buffer)
+  "Serialize OBJECT with message descriptor MSG-DESC into BUFFER using wire format.
+   The value returned is the number of octets written to BUFFER."
   (declare (buffer buffer))
```
cgay, line 108: Could you add a type decl for msg-desc since the method had one?

bkuehnert: Good idea. Done

PR #116, opened 2020-07-27
Update CCL-test.yml

Slids:


Slids:
Note: The ccl-tests will fail until the 
https://github.com/qitab/cl-protobufs/pull/114 and
https://github.com/qitab/cl-protobufs/pull/113
are committed

PR #122, opened 2020-07-27
Remove dead lazy-reader code

bkuehnert:
This code is redundant, as we make the lazy reader in
make-structure-class-forms-lazy. Since moving to structs rather than
classes, this code is totally dead.

The proto-%bytes utility function survived, since it sees some use
in the testing function (and I'd bet that QPX uses it). However,
it should probably be moved somewhere else.


Slids:
proto-%bytes is not used in qpx

PR #126, opened 2020-07-28
Fix some abcl warnings

bkuehnert:
ABCL caught these unused variables. They are indeed unused, so they are being removed.


Slids:
With this we get a new warning/failure:
ERROR: FAIL: SB-INT:SIMPLE-STYLE-WARNING 'The variable PROTOBUFS-IMPLEMENTATION::IDX is defined but never used.'

PR #127, opened 2020-07-29
Fix two lazy (de)serialization bugs

bkuehnert:
This PR fixes the following two bugs:

1. Repeated lazy fields were not being properly serialized by `emit-repeated-field`. The `%bytes` slot was being ignored, and the default value for the slot was being serialized (so, always `nil`). To fix this, I added a check to the top of `serialize-object` to see if the message in question has a bound `%bytes` slot. If it does, then just print that to buffer. This cuts down on repeated code (since previously, this check for a `%bytes` slot would need to happen in `emit-non-repeated-field` AND `emit-repeated-field`). Someone did write a test for this problem, but the test wasn't working properly. The test is supposed to deserialize lazily and then immidiately serialize this proto (so the lazily deserialized bytes are immediately written to the wire). However, in between deserializing and serializing, the accessors were called. So, at that step the lazy fields were deserialized prematurely.

2. The optimized (de)serializers were ignoring fields' being declared lazy. Now, the optimized serializers will output the `%bytes` slot to the wire if bound, and the optimized deserializers will call `make-message-with-bytes` if `proto-lazy-p` is set for a field. I've also edited the lazy tests so that they properly test the optimzed (de)serializers.


serialize.lisp
```diff
@@ -107,20 +107,32 @@
...
-             (data      (slot-value object (oneof-descriptor-internal-name oneof)))
-             (set-field (oneof-set-field data))
-             (value     (oneof-value data)))
-        (when set-field
-          (let* ((field (aref fields set-field))
-                 (type  (proto-class field))
-                 (index (proto-index field)))
-            (iincf size
-                   (emit-non-repeated-field value type index buffer))))))))
+  (let ((bytes (and (slot-exists-p object '%bytes)
```
cgay, line 110: I realize this happens elsewhere in our code so you're just following the precedent, but we ought to know whether this slot exists or not so the fact that we're asking at run time feels like lazy programming.  Every message is a substruct of `base-message`, and every base-message has a `%bytes` slot.  Worst case we should put `(if (not (typep object 'base-message)) (error "..."))`.  Can we instead just put `(declare (type base-message object))` and omit the `slot-exists-p` check?

cgay: Slight mistake there. Not every `base-message` has a `%bytes` slot, but every generated message does. (So effectively every `base-message`.)

bkuehnert: I believe that this check is done to preclude groups. Groups are the oddball since groups are substructs of `base-message` and yet they do not have a `%bytes` slot. 

A simple way to get around this is to just declare that the type of `object` to be the `proto-class` slot of `msg-desc`. Then we know for sure that this is a message object for sure.

bkuehnert: Update: My "simple" solution won't work. Since groups also use `message-descriptors`, then this type declaration does nothing. 

Additionally: the function `serialize-object` turns out to be called with `object` bound to a group. So, this check is needed.

How about just a comment that says "We check for the %bytes slot, since groups do not have this slot"?

cgay: That would be great, yes.

serialize.lisp
```diff
@@ -977,16 +982,29 @@ Parameters:
...
+          (let ((,bytes (and (slot-exists-p ,vobj '%bytes)
+                             (proto-%bytes ,vobj))))
+            ;; If BYTES is bound, then VOBJ is a lazy field, and BYTES is the pre-computed
+            ;; serialization of VOBJ. So, just output that.
+            (cond
+              (,bytes
+               (setf ,size (length ,bytes))
+               (buffer-ensure-space ,vbuf ,size)
+               (fast-octets-out ,vbuf ,bytes)
+               ,size)
```
Slids, line 1004: no new line here

serialize.lisp
```diff
@@ -977,16 +982,29 @@ Parameters:
...
       (def-pseudo-method :serialize message `(,vobj ,vbuf &aux (,size 0))
         `((declare ,$optimize-serialization)
           (declare (ignorable ,vobj ,vbuf))
           (declare ; maybe allow specification of the type
            #+ignore(type ,(proto-class message) ,vobj)
            (type fixnum ,size))
-          ,@serializers
-          ,size)))))
+          ;; Check for the %BYTES slot, since groups do not have this slot.
+          (let ((,bytes (and (slot-exists-p ,vobj '%bytes)
```
Slids, line 995: slot-exists-p: why? The slot will always exist right?

bkuehnert: See the comment. Groups do not have this slot, and group lazy (de)serialization isn't supported

Slids:
The cla check was def wrong...

PR #128, opened 2020-07-29
Modernize ASDF style

sionescu:
Move ASDF component classes to a separate system, and use
defsystem-depends-on to load that.
Move test defsystem to the main .asd file.


cl-protobufs.asd
```diff
@@ -244,4 +97,211 @@ which means ASDF loads both the .lisp file and the .fasl file."
...
+                            #:cl-protobufs.test.map-test
+                            #:cl-protobufs.test.oneof-test
+                            #:cl-protobufs.test.import-test
+                            #:cl-protobufs.test.lazy-test
+                            #:cl-protobufs.test.alias-test
+                            #:cl-protobufs.test.packed-test
+                            #:cl-protobufs.test.serialize-test
+                            #:cl-protobufs.test.text-format-test
+                            #:cl-protobufs.test.zigzag-test
+                            #:cl-protobufs.test.well-known-types-test))))
```
cgay, line 307: Why is it necessary to list all the tests here? We should avoid this if at all possible because it will get out of sync with the test code, especially given that most dev (currently) happens using Bazel, not ASDF.

sionescu: Because the current code uses a separate package for each test suite (take a look at the code this is replacing).

bkuehnert: This is because our way of writing testing functions is ad-hoc. Each test file named `*-test` happens to define a package `cl-protobufs.test.*-test` which exports a function `run` that runs the relevant tests. I suppose we could make an ASDF component that loads .lisp files (assuming they are written in the above way) and then adds the symbol for the `run` function to some global list. That seems ugly, but atleast we wouldn't have to update code in two places when making a change.

sionescu: That would be pretty error-prone. The simple solution is to have all tests share one package.

cgay: Maybe this is the best we can do with the clunit2, I don't know.  In general it should be possible to have `deftest` record each test, then load the `:cl-protobufs/tests` system and tell the unit test infrastructure to run all the tests. Then it doesn't matter what package they're in and there's no need to maintain a separate list.  That's for another day though...

cl-protobufs.asdf.asd
```diff
@@ -0,0 +1,13 @@
+;;; Copyright 2012-2020 Google LLC
+;;;
+;;; Use of this source code is governed by an MIT-style
+;;; license that can be found in the LICENSE file or at
+;;; https://opensource.org/licenses/MIT.
+
+(defsystem :cl-protobufs.asdf
+  :description "ASDF component classes for cl-protobufs."
+  :author "Ben Kuehnert <benkuehnert@google.com>"
```
bkuehnert, line 9: I think Robert Brown should be credited for writing the ASDF utility file. I only made minor changes to his .asd file: https://github.com/brown/protobuf/blob/master/protobuf.asd.

sionescu: Fixed.

bkuehnert: Thanks!

Slids:
Lint:
CheckLint/0: Warning (//depot/google3/third_party/lisp/METADATA:17:3)
CheckLint found errors.
Don't add new copyright statements. (COPYRIGHT)
	//depot/google3/third_party/lisp/cl_protobufs/asdf.lisp:1
No mention of parameter COMPONENT in docstring for function COMPONENT-DEPENDS-ON. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/asdf.lisp:51
No mention of parameter COMPONENT in docstring for function COMPONENT-DEPENDS-ON. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/asdf.lisp:58
No mention of parameter PROTOBUF-SOURCE-FILE in docstring for function PROTO-INPUT. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/asdf.lisp:65
No mention of parameter COMPONENT in docstring for function OUTPUT-FILES. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/asdf.lisp:82
No docstring for function RESOLVE-SEARCH-PATH. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/asdf.lisp:101

Slids:
New lint:
CheckLint/0: Warning (//depot/google3/third_party/lisp/METADATA:17:3)
CheckLint found errors.
No mention of parameter COMPONENT in docstring for function COMPONENT-DEPENDS-ON. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/asdf.lisp:59
No mention of parameter COMPONENT in docstring for function OUTPUT-FILES. (MISSING-DOCUMENTATION)
	//depot/google3/third_party/lisp/cl_protobufs/asdf.lisp:84

PR #133, opened 2020-07-31
Add protoc binary to cl-protobufs

Slids:
Both add the binary and update the readme.
We change asdf to use the added binary for x86-64.
We should also make a github action to generate these...


Slids:
I'm the only committer, the cla check is wrong...

sionescu:
Hi @Slids, you should squash all these commits. I think that's what confuses the bot. And please use [LFS](https://docs.github.com/en/github/managing-large-files/configuring-git-large-file-storage).

README.md
```diff
@@ -46,6 +46,10 @@ Common Lisp.
 
 ## Using `protoc` to Generate Lisp Code
 
+If you are running on x86-64 on linux we have bundled the proto-gen-lisp
+binary with cl-protobufs. You will have to install protoc,
+which can be done with apt-get. This is best effort.
```
bkuehnert, line 51: "This is best effort." What is meant by this?

PR #134, opened 2020-07-31
Use first enum as default rather than lowest index

bkuehnert:
Proto2 AND proto3 specify that the first enum is always the default.
This code previously used the lowest index enum as the default, which
is incorrect.

### DIFF
--- a/define-proto.lisp
+++ b/define-proto.lisp
@@ -317,17 +317,10 @@ return nil."
 
 (defun make-enum-default (type enum-values)
   "Generate a function to return the default enum value for
-an enum of type TYPE. The default type should be the enum
-with the lowest index value in ENUM-VALUES."
-  (let ((default-value))
-    (loop with smallest-value = nil
-          for enum in enum-values
-          when (or (not smallest-value)
-                   (< (enum-value-descriptor-value enum) smallest-value))
-            do (setf smallest-value (enum-value-descriptor-value enum)
-                     default-value (enum-value-descriptor-name enum)))
-    `(defmethod enum-default-value ((e (eql ',type)))
-       ,default-value)))
+an enum of type TYPE. The default type should be the first
+enum in ENUM-VALUES."
+  `(defmethod enum-default-value ((e (eql ',type)))
+     ,(enum-value-descriptor-name (car enum-values))))
 
 (defun make-enum-constant-forms (type enum-values)
   "Generates forms for defining a constant for each enum value in ENUM-VALUES.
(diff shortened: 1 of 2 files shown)


define-proto.lisp
```diff
@@ -317,17 +317,10 @@ return nil."
...
-          when (or (not smallest-value)
-                   (< (enum-value-descriptor-value enum) smallest-value))
-            do (setf smallest-value (enum-value-descriptor-value enum)
-                     default-value (enum-value-descriptor-name enum)))
-    `(defmethod enum-default-value ((e (eql ',type)))
-       ,default-value)))
+an enum of type TYPE. The default type should be the first
+enum in ENUM-VALUES."
+  `(defmethod enum-default-value ((e (eql ',type)))
+     ,(enum-value-descriptor-name (car enum-values))))
```
cgay, line 323: How about just inlining this in the one place it's used.

Slids: Note: The defmethod has a generic for when this has yet to be evaluated
It's needed for nested enums.
Try removing it, you'll get an error...

cgay: There is no `defgeneric` for this method and if there were one, or if we're depending on the implicitly defined one, how does a generic with no methods help anything? Possibly I'm missing your point.

cgay: nevermind, i found it.

cgay: Note that I said nothing about removing the defmethod.

define-proto.lisp
```diff
@@ -317,17 +317,10 @@ return nil."
...
-  (let ((default-value))
-    (loop with smallest-value = nil
-          for enum in enum-values
-          when (or (not smallest-value)
-                   (< (enum-value-descriptor-value enum) smallest-value))
-            do (setf smallest-value (enum-value-descriptor-value enum)
-                     default-value (enum-value-descriptor-name enum)))
-    `(defmethod enum-default-value ((e (eql ',type)))
-       ,default-value)))
+an enum of type TYPE. The default type should be the first
```
cgay, line 320: "The default type" -> "The default value"

(Or inline this whole function and then you don't have to fix the doc.)

define-proto.lisp
Slids: Then I have no idea what your talking about...

bkuehnert: It seems like a useful function to have around, although yes it seems eclipsed by the `get-default-form` function.

cgay: It's a function. I suggested inlining it because it's two lines and only called in one place. It's not a big deal but I'm not sure why that's hard to understand.  Anyhoo...moving on.

bkuehnert:
merged with https://github.com/qitab/cl-protobufs/commit/b8ba39994a76d4dfb1166e5eb799744f18e6fffc

PR #143, opened 2020-08-04
Proto3 Syntax

bkuehnert:
This PR adds support for proto3 syntax. Proto3 singular fields are supported, as well as the experimental proto3 optional fields.

Some things of note
1. There is a chunk of code duplication between `serialize-object` and `print-text-format`. In both of these functions, I needed to add a special condition for singular fields to ensure that we don't serialize/print the default values. This can be avoided by calling the has-* function here (which is what the optimized serializer does). The problem with this, is that we need to look up the name of the has-* function in `*proto-function-table*`, which is possibly slow.
2. With Carl's new PR, the `*syntax*` global variable is redundant. I will merge his branch in so that we use that to determine the current file's syntax. (fixed)
3. This is silly, but line 127 of asdf.lisp goes over the column limit and I wasn't able to break the line without making the string two lines. I tried adding a backslash to cancel the linebreak, but that didn't work. I know there is some simple way to do it, I just couldn't figure out the syntax. (fixed)


asdf.lisp
```diff
@@ -124,7 +124,7 @@ to PARENT-PATH."
          ;; that method instead of executing (component-pathname component).
          (output-file (first (output-files operation component)))
          (search-path (cons (directory-namestring source-file) (resolve-search-path component)))
-         (command (format nil "protoc --proto_path=~{~A~^:~} --lisp_out=output-file=~A:~A ~A"
+         (command (format nil "protoc --proto_path=~{~A~^:~} --lisp_out=output-file=~A:~A ~A --experimental_allow_proto3_optional"
```
Slids, line 127: Why do we need this command line?
Please remove the flag when we're ready to commit.

define-proto.lisp
```diff
@@ -503,6 +505,7 @@ Parameters:
        ,(make-oneof-descriptor
          :internal-name internal-name
          :external-name name
+         :synthetic-p (and synthetic-p t)
```
Slids, line 508: Why the and?

define-proto.lisp
```diff
@@ -553,21 +555,38 @@ Arguments:
...
-          (= (bit (,is-set-accessor ,obj) ,index) 1))
+          ,(if index
+               `(= (bit (,is-set-accessor ,obj) ,index) 1)
+               `(let ((,cur-value ,(if bool-index
+                                       `(plusp (bit (,bit-field-name ,obj) ,bool-index))
+                                       `(,hidden-accessor-name ,obj))))
+                  ,(case (proto-set-type field)
+                     ((proto:byte-vector cl:string) `(not (= (length ,cur-value) 0)))
+                     ((cl:double-float cl:float) `(not (= ,cur-value ,default-form)))
+                     (t `(not (eq ,cur-value ,default-form)))))))
```
Slids, line 589: What is the possibilities for t?
Comment maybe?

define-proto.lisp
```diff
@@ -593,7 +613,12 @@ Arguments:
         ;; field, this is impossible.
         (proto-impl::set-functions-in-hash-table ',proto-type ',public-slot-name)
 
-        (export '(,has-function-name ,clear-function-name ,public-accessor-name))))))
+        ;; has-* functions are not exported for singular fields. They are only for
+        ;; internal usage.
+        ,(when (not (eq label :singular))
```
Slids, line 618: ,(unless (eq label :singular)

model-classes.lisp
```diff
@@ -102,10 +102,10 @@ Parameters:
 
 ;; A Protobufs schema, corresponds to one .proto file
 (defclass file-descriptor (descriptor)
-  ((syntax :type (or null string)               ; syntax, passed on but otherwise ignored
+  ((syntax :type keyword               ; proto syntax. Either :proto2 or :proto3.
```
Slids, line 105: :type (or :proto2 proto3)

model-classes.lisp
```diff
@@ -102,10 +102,10 @@ Parameters:
...
            :accessor proto-syntax
            :initarg :syntax
-           :initform "proto2")
+           :initform :proto2)
```
Slids, line 108: Seems strange, should we do this?

protoc/generator.h
```diff
@@ -39,6 +39,11 @@ class LispGenerator : public compiler::CodeGenerator {
                    compiler::GeneratorContext* generator_context,
                    std::string* error) const override;
 
+  // cl-protobufs supports proto3 optional fields.
+  uint64_t GetSupportedFeatures() const override {
+    return FEATURE_PROTO3_OPTIONAL;
+  }
```
Slids, line 45: remove this before submitting.

asdf.lisp
bkuehnert: https://github.com/protocolbuffers/protobuf/blob/9952e36f92cfff9ca0ec4d43a5ff146d26f64a01/docs/implementing_proto3_presence.md

The flag is currently required if you want protoc to not throw an error when seeing proto3 optional fields. I have tried removing it, it doesn't work.

model-classes.lisp
bkuehnert: Maybe not. This is probably a relic from manual schema definition. I'm guessing someone got tired of writing ":syntax "proto2"" when it did literally nothing.

tests/proto3-test.lisp
```diff
@@ -0,0 +1,104 @@
...
+        (assert-true (proto-impl::proto-equal deserialized msg))))))
+
+(deftest optional-test (proto3-tests)
+  (let ((msg (make-optional-test)))
+    (assert-false (optional-test.has-bool-value msg))
+    (assert-false (optional-test.has-string-value msg))
+    (setf (string-value msg) "")
+    (setf (bool-value msg) nil)
+    (assert-true (optional-test.has-bool-value msg))
+    (assert-true (optional-test.has-string-value msg))))
```
Slids, line 104: Add another mediocre performance test?

tests/proto3.proto
```diff
@@ -0,0 +1,37 @@
...
+}
+
+message TestMessage {
+  int32 value = 1;
+}
+
+message OptionalTest {
+  optional string string_value = 1;
+  optional bool bool_value = 2;
+}
```
Slids, line 37: Newline and add one with some optional and some singular?

bkuehnert:
> IS there a good way to deduplicate the mentioned duplication?

Other than just calling the `has-*` function? We can always make a new inlined function, but that doesn't seem great

define-proto.lisp
bkuehnert: I was getting an error that looked like "T is not of type (MEMBER NIL T)", since the slot is typed as a boolean. I'm guessing this is because this function is a macro, and the value of SYNTHETIC-P is 'T rather than T? Anyway, it follows the convention of what's done for the LAZY-P argument in the other macros.

asdf.lisp
Slids: I see, it's still being implemented in many protobuf implementations /sigh, will need to make a bug with the protobuf team to notify when to remove it...

Slids:
> > IS there a good way to deduplicate the mentioned duplication?
> 
> Other than just calling the `has-*` function? We can always make a new inlined function, but that doesn't seem great

I don't know, maybe just a function

(declare (inline set-to-default))
(defun set-to-default (type value)
  (case type
            ((proto:byte-vector cl:string) 
             (= (length value) 0))
            ((cl:double-float cl:float) 
             (= value (get-default-form type)
                           (proto-default field))))
                 (t (eq value (get-default-form type))))

It's used in a few places and seems valid enough

model-classes.lisp
Slids: Since it does something now we should probably remove it.
I don't see a case where this could validly not be set by a constructor...

bkuehnert:
> > > IS there a good way to deduplicate the mentioned duplication?
> > 
> > 
> > Other than just calling the `has-*` function? We can always make a new inlined function, but that doesn't seem great
> 
> I don't know, maybe just a function
> 
> ```
> (declare (inline set-to-default))
> (defun set-to-default-p (type value)
>   (case type
>             ((proto:byte-vector cl:string) 
>              (= (length value) 0))
>             ((cl:double-float cl:float) 
>              (= value (get-default-value type))))
>             (t (eq value (get-default-value type))))
> ```
> 
> It's used in a few places and seems valid enough

Another benefit of using the `has-*` function: the call to `case` is already computed at compile time.

Slids:
Crazy idea would be to store the has-function (and maybe get-function) name on the symbol plist...
Then we could remove those hash-tables...
Probably a different pr though...

model-classes.lisp
```diff
@@ -733,6 +733,8 @@ on the symbol if we are not in SBCL."
 
 (defstruct oneof-descriptor
   "The meta-object for a protobuf oneof"
+  ;; A boolean which indicates if the oneof is synthetic.
```
cgay, line 736: This is a good place to define the term "synthetic".

define-proto.lisp
bkuehnert: Updating with a comment. Non-string non-float scalars are sometimes referred to as "integral types" in proto docs.

model-classes.lisp
bkuehnert: But what about :proto4!!! /s

asdf.lisp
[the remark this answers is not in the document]
bkuehnert: error 403 :(

cgay: http://www.lispworks.com/documentation/HyperSpec/Body/22_cic.htm

Slids:
You were asking about tests passing but the code failing:
https://github.com/qitab/cl-protobufs/runs/946566394?check_suite_focus=true

tests/proto3-test.lisp
bkuehnert: How about just turning the serialization test into a perf test (by running in 5000+ times)?

Slids: It should have some level of nesting, and multiple protos and be of decent size, can we copy the one in serialize test?

bkuehnert:
> You were asking about tests passing but the code failing:
> https://github.com/qitab/cl-protobufs/runs/946566394?check_suite_focus=true
> 
> Edit:
> The error is due to my ask to remove the initform, we found it why it was needed!
> But it probably should init to something thats not :proto2 or :proto3, like nil...

I fixed the error for SBCL, although I never got an error locally. It seems like CCL is still having problems.
Why should it default to NIL? What is the case where we would be correct to not set the syntax?

Slids:
Never, but setting the initform to :proto2 or :proto3 when we don't know
yet is wrong.

Slids:
Only one lint failure:
CheckLint found errors.
Lines should be <= 80 columns wide  [whitespace/line_length] [2]
	//depot/google3/third_party/lisp/cl_protobufs/protoc/message.cc:150

Slids:
Also please remove:
proto3.proto~
proto3-test.lisp~

bkuehnert:
> Also please remove:
> proto3.proto~
> proto3-test.lisp~
> 
> Also please update the run functions in proto3-test.lisp
> a) for carls changes
> b) you have the wrong test suite name
> 
> And:
> third_party/lisp/cl_protobufs/protoc/file.cc:48:11: error: enumeration value 'SYNTAX_UNKNOWN' not handled in switch [-Werror,-Wswitch]
> switch (file_->syntax()) {

Wow, no lisp linting errors! Should be all fixed now.

I know the `run` function was wrong, but I think I already fixed the rest of it. The package name and suite name have already been changed to match Carl's changes.

PR #144, opened 2020-08-04
Fix the length and nbytes written for uint64-integer-decoders

Slids:
Currently we find the uint32 length for uint64 encoders.
This causes 3 test failures. 
This fixes it.


Slids:
I'm sure I am me, I work at Google, and my username is attached, so wtf?

Slids:
> Could you squash the two commits together, please?

I did...
Doesn't seem like it though...

Slids:
Okay, done, and the cla is fine...

PR #145, opened 2020-08-04
Update README for proto3

bkuehnert:


README.md
```diff
@@ -243,6 +245,32 @@ Clear the value of a field. After the above call `(date-range.has-min-date
...
+}
+```
+
+The generated code for proto3 messages is similar to proto2 messages. The only
+difference is the introduction of fields with no specified label, which are
+known as "singular" fields. For singular fields, the state of being unset
+and the state of being set to the default value for the type are indistinguishable.
+So, `has-*` functions, such as `(event.has-day msg)` are not defined.
+
+The `has-*` functions for repeated fields are defined, and work as normal in proto3.
```
Slids, line 269: Not a huge fan of "work as normal", should explain normal.

README.md
```diff
@@ -243,6 +245,32 @@ Clear the value of a field. After the above call `(date-range.has-min-date
...
+
+This library supports optional fields in proto3 messages. These fields have
+the same semantics as proto2 optional fields (namely, they have `has-*` functions).
```
Slids, line 272: Remove  (namely, they have `has-*` functions)?
Does it add anything?

README.md
```diff
@@ -366,7 +394,8 @@ message Person {
   }
 }
 ```
-To access fields inside a oneof, just use the standard accessors outlined above. For example:
+To access fields inside a oneof, just use the standard accessors outlined above. These fields
```
Slids, line 397: remove just
In all honesty I probably would have written it to...

README.md
bkuehnert: No, the fields are identical to proto2 optional fields in terms of generated code.

cgay:
made these changes internally. closing.

PR #150, opened 2020-08-06
Remove some dead clos-transform code

bkuehnert:
This code was probably used for parsing protos with lisp, but it is no
longer used. The protoc plugin is able to convert from proto custom
defaults to lisp style custom defaults.

note: the entire clos-transform.lisp file is redundant by protoc/literals.cc. I'm currently trying to refactor the code so that it can be safely deleted.

### DIFF
--- a/clos-transform.lisp
+++ b/clos-transform.lisp
@@ -209,38 +209,3 @@ https://developers.google.com/protocol-buffers/docs/proto#scalar "
   (not (null (member type '(:int32 :int64 :uint32 :uint64 :sint32 :sint64
                             :fixed32 :fixed64 :sfixed32 :sfixed64
                             :bool :float :double)))))
-
-(defun protobuf-default-to-clos-init (default type)
-  "Given a Protobufs type and default, return a CLOS initform value.
-   Don't call this if the default is empty, because that will confuse 'nil' with 'unbound'."
-  (cond ((ignore-errors (typep default type))
-         default)
-        ((symbolp default)
-         (cond ((eq type :bool)
-                (boolean-true-p default))
-               ;; If we've got a symbol, it must be to initialize an enum type
-               ;; whose values are represented by keywords in Lisp
-               (t (kintern (symbol-name default)))))
-        ((stringp default)
-         (cond ((eq type :bool)
-                (boolean-true-p default))
-               ((member type '(:int32 :uint32 :int64 :uint64 :sint32 :sint64
-                               :fixed32 :sfixed32 :fixed64 :sfixed64))
-                (let ((default (read-from-string default)))
-                  (and (integerp default) default)))
-               ((member type '(:float :double))
-                (let ((default (read-from-string default)))
-                  (and (floatp default) default)))
-               (t default)))))
-
-(defun boolean-true-p (x)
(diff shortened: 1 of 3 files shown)


clos-transform.lisp
```diff
@@ -209,38 +209,3 @@ https://developers.google.com/protocol-buffers/docs/proto#scalar "
...
-               ((member type '(:int32 :uint32 :int64 :uint64 :sint32 :sint64
-                               :fixed32 :sfixed32 :fixed64 :sfixed64))
-                (let ((default (read-from-string default)))
-                  (and (integerp default) default)))
-               ((member type '(:float :double))
-                (let ((default (read-from-string default)))
-                  (and (floatp default) default)))
-               (t default)))))
-
-(defun boolean-true-p (x)
```
Slids, line 236: This is a wonderful function...

text-format.lisp
```diff
@@ -323,7 +323,11 @@ return T as a second value."
              ((:float) (parse-float stream))
              ((:double) (parse-double stream))
              ((:string) (parse-string stream))
-             ((:bool)   (boolean-true-p (parse-token stream)))
+             ((:bool)   (let ((token (parse-token stream)))
+                          (cond ((string= token "true") t)
+                                ((string= token "false") nil)
+                                ;; signal an error
```
Slids, line 329: You say signal an error, but this is not a call to signal.
Do you want to signal an error or revise this comment?

bkuehnert: The function returns T as a second value if parsing fails. This is so that the we can signal an error with the msg-desc and field-desc at the call site, since those variables aren't accessible in this helper function. I'll update the comment to make this more clear.

clos-transform.lisp
bkuehnert: "`x` isn't "YES" or "TRUE"? Ok, guess it's NIL then"

text-format.lisp
```diff
@@ -323,7 +323,12 @@ return T as a second value."
              ((:float) (parse-float stream))
              ((:double) (parse-double stream))
              ((:string) (parse-string stream))
-             ((:bool)   (boolean-true-p (parse-token stream)))
+             ((:bool)   (let ((token (parse-token stream)))
+                          (cond ((string= token "true") t)
+                                ((string= token "false") nil)
+                                ;; Parsing failed, return T as
+                                ;; a second value to signal an error.
```
Slids, line 330: I still don't like the documentation.
You should write the documenation from the functions point of view,
any expectation of what the callsite does is 
a) flaky
b) likely to someday be wrong.

bkuehnert: I want to say that the function is signalling that it had a problem parsing. I do not mean to imply that any lisp error condition is being triggered. The term "signal an error" is overloaded...

Slids: In lisp it has a very technical meaning:
http://www.ai.mit.edu/projects/iiip/doc/CommonLISP/HyperSpec/Body/fun_signal.html
Have to be careful with it :)

bkuehnert: I had to consult a thesaurus, but I've updated the comment again (and added a similar comment at the end of the main `cond`).

PR #156, opened 2020-08-11
Refactor field-descriptor

bkuehnert:
This PR makes the following changes to the field-descriptor class.

1. The TYPE slot has been repurposed. Instead of storing a string that named the protobuf type, it now stores the base lisp type for the field. "Base" here means that if a type is repeated, then TYPE is bound to [type] rather than (list-of [type]). This is because this information is duplicated by the label and container fields anyway.
2. The CLASS slot has been repurposed. Instead of storing the lisp type (if it is an enum, map, or message) or a keyword (if it's scalar), it now stores a keyword which indicates if the type is an enum, map, message, or scalar. This is useful since we often need to distinguish between these different classes of types, and previously this was impossible at compile time.
3. DEFAULT has been removed and replaced with INITFORM. Previously default held either a default or a symbol which indicated if the default was an empty list or vector. This was confusing and unnecessary, and it required re-computing a type's default several times.
5. CONTAINER has been added as a slot that stores either :LIST, :VECTOR, or NIL. This indicates which type of lisp container is used for a repeated field (or NIL if the field is not repeated). This information is required by the deserializer, and was previously pulled out of the DEFAULT slot. 
6. SET-TYPE has been removed. It previously stored the lisp type.
7. LISP-TYPE has been removed. It didn't do anything with the removal of aliased types.

Apologies in advance for the giant PR.


api.lisp
```diff
@@ -20,10 +20,10 @@ The definition of initialized is all required-fields are set."
         when (= (bit (slot-value object '%%is-set)
                      (proto-field-offset field))
                 1)
-          do (let ((lisp-type (proto-class field))
+          do (let ((lisp-type (proto-type field))
```
Slids, line 23: pull this let into the loop with with foo = ...?

api.lisp
```diff
@@ -20,10 +20,10 @@ The definition of initialized is all required-fields are set."
...
+                   (proto-class (proto-class field))
                    (field-value (slot-value object (proto-internal-field-name field))))
-               (when (and (not (keywordp lisp-type))
-                          (find-message lisp-type))
+               (when (member proto-class '(:message :group))
```
Slids, line 26: then pull when into the loop decleration?

define-proto.lisp
```diff
@@ -471,23 +470,19 @@ Parameters:
...
+           (slot &key type class name (default nil default-p)
                  lazy index documentation &allow-other-keys)
            field
          (assert index)
-         (let ((default (if default-p default $empty-default)))
-           (multiple-value-bind (ptype pclass packed-p enum-values root-lisp-type)
-               (clos-type-to-protobuf-type type)
+         (let* ((default (when default-p default))
+                (initform (get-default-form type class :optional default)))
              (declare (ignore packed-p enum-values))
```
Slids, line 479: I think this is here due to the multiple value bind, can we delete it now?

define-proto.lisp
```diff
@@ -899,30 +896,25 @@ Arguments:
...
+        ((eq class :scalar)
+         (gethash type default-form))
+        ((eq class :map)
          '(make-hash-table))
-        ((enum-default-value `,type) (enum-default-value `,type))
-        (t `(enum-default-value ',type))))))
+        ((eq class :enum)
+         `(enum-default-value ',type))
+        ((member class '(:group :message))
+         nil)))))
```
Slids, line 917: add a default case that signals an error?

serialize.lisp
```diff
@@ -244,79 +246,82 @@ Parameters:
...
-           (cond ((not value) 0)
-                 ((eq (proto-message-type desc) :group)
-                  (let ((tag1 (make-wire-tag $wire-type-start-group index))
-                        (tag2 (make-wire-tag $wire-type-end-group   index)))
-                    (iincf size (encode-uint32 tag1 buffer))
-                    (dolist (f (proto-fields desc)
-                               (i+ size (encode-uint32 tag2 buffer)))
-                      (iincf size (emit-field value f buffer)))))
+    (case class
+      ((:scalar) (serialize-scalar value type (make-tag type index)
```
Slids, line 265: maybe new line?

text-format.lisp
```diff
@@ -310,81 +312,83 @@ attempt to parse the name of the message and match it against MSG-DESC."
...
-                 (t
-                  (skip-whitespace stream)
-                  (list (parse-map-entry key-type val-type stream)))))))
-          ;; Parsing failed, return t as a second vlaue to indicate failure.
-          (t (values nil t)))))
+(defun parse-field (type class &key (stream *standard-input*))
+  "Parse data of type TYPE and class CLASS from STREAM. This
+function returns the object parsed. If the parsing fails,
+the function will return T as a second value."
+  (let ((desc))
```
Slids, line 327: What a horrid name for a variable.
What is it?
Comment please?
(I know, already there..)

Slids:
Great work, made a few comments

api.lisp
bkuehnert: Looking at this function again, it appears to be bugged. If MESSAGE has an optional boolean field, then this function will fail. It should be rewritten to assert that the field is a message type before any of these let bindings happen. In that case, we won't be able to move the let into the loop.

bkuehnert: Fixed. Resolving this convo, feel free to open another one on the new change.

define-proto.lisp
bkuehnert: You're right. Same thing happens in `define-oneof`. Removing

define-proto.lisp
bkuehnert: Done.

I've also added an error for the case where a custom type doesn't have an explicitly set default. Previously, unknown types would always have a NIL default, but that should really throw an error.

serialize.lisp
```diff
@@ -776,112 +788,116 @@ Parameters:
...
+            (let ((,vval ,reader))
+              (flet ((serialize-pair (k v)
+                       (let ((ret-len (encode-uint32 ,tag ,vbuf))
+                             (map-len 0))
+                         (with-placeholder (,vbuf)
+                           (iincf map-len (serialize-scalar k ,key-type
+                                                            ,(make-tag key-type 1)
+                                                            ,vbuf))
+                           ,(generate-non-repeated-field-serializer
+                             val-type val-class 2 'v 'v vbuf 'map-len)
```
cgay, line 896: I'm unsure how this works. Since `v` is used as the value for the `boundp` arg here, doesn't this mean that if `v` is nil that only the key will be serialized and the corresponding value won't?  Is that ok?

text-format.lisp
```diff
@@ -310,81 +312,85 @@ attempt to parse the name of the message and match it against MSG-DESC."
...
-                             (skip-whitespace stream)
-                             (expect-char stream #\])
-                             (return pairs)))))
-                 (t
-                  (skip-whitespace stream)
-                  (list (parse-map-entry key-type val-type stream)))))))
-          ;; Parsing failed, return t as a second vlaue to indicate failure.
-          (t (values nil t)))))
+(defun parse-field (type class &key (stream *standard-input*))
+  "Parse data of type TYPE and class CLASS from STREAM. This
```
cgay, line 324: "class CLASS" should be "keyword CLASS" here, but I hope you'll rename "class" to "category" (or something better) throughout.

text-format.lisp
```diff
@@ -310,81 +312,85 @@ attempt to parse the name of the message and match it against MSG-DESC."
...
+function returns the object parsed. If the parsing fails,
+the function will return T as a second value."
+  ;; placeholder descriptor. This may be bound to a descriptor for a
+  ;; message, group, map, or enum.
+  (let ((descriptor))
```
cgay, line 329: Same thing here. Don't create a binding with a huge scope and then setq it, unless absolutely necessary.

text-format.lisp
```diff
@@ -310,81 +312,85 @@ attempt to parse the name of the message and match it against MSG-DESC."
...
+       (let ((key-type (map-descriptor-key-type descriptor))
+             (val-type (map-descriptor-val-type descriptor))
+             (val-class (map-descriptor-val-class descriptor)))
+         (flet ((parse-map-entry (key-type val-type stream)
+                  (let (key val)
+                    (expect-char stream #\{)
+                    (assert (string= "key" (parse-token stream)))
+                    (setf key (parse-field key-type :scalar :stream stream))
+                    (skip-whitespace stream)
+                    (assert (string= "value" (parse-token stream)))
```
cgay, line 370: These assertions are going to produce error messages without any context (filename, line, column). We should fix that at some point.

wire-format.lisp
```diff
@@ -177,71 +177,71 @@ Parameters:
...
                        (string val)
                        ;; Non-keyword symbols are consy, avoid them if possible
                        (format nil "~A:~A" (package-name (symbol-package val)) (symbol-name val)))))
           (encode-string val buffer)))
-       ((:date :time :datetime :timestamp)
+       ((date time datetime timestamp)
         (encode-int64 val buffer))))))
 
 (defun get-scalar-encoder-form (type val buffer)
   "Returns a form that encodes a value VAL with type
```
cgay, line 207: I mean, no, VAL is not a value to be encoded, it's a symbol naming a variable to be bound to a value to be encoded. But I guess that's a mouthful.

model-classes.lisp
```diff
@@ -309,12 +309,10 @@ Parameters:
...
   "The meta-object for a protobuf map"
-  (class     nil :type symbol)
   (name      nil :type string)
-  (key-class nil :type symbol) ;; the :class of the key
-  (val-class nil :type symbol) ;; the :class of the value
-  (key-type nil :type symbol)  ;; the lisp type of the key
-  (val-type nil :type symbol)) ;; the lisp type of the value
+  (key-type  nil :type symbol)                                  ; the lisp type of the key
+  (val-type  nil :type symbol)                                  ; the lisp type of the value
+  (val-class nil :type (member :scalar :message :group :enum))) ; the class of the value
```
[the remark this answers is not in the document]
Slids: Probably not, not used to often and we'd have to got find the different types.

model-classes.lisp
```diff
@@ -513,9 +505,13 @@ on the symbol if we are not in SBCL."
            :accessor proto-writer               ; it's a list, it's something like '(setf title)'
            :initarg :writer
            :initform nil)
-   (default :accessor proto-default             ; Default value (untyped), pulled out of the options
-            :initarg :default
-            :initform $empty-default)
+   (initform :accessor proto-initform           ; Default initform
```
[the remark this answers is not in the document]
Slids: Is this actionable or just a note?

serialize.lisp
```diff
@@ -155,27 +157,27 @@ Parameters:
...
       (when
           (and (eq (proto-label field) :singular)
-               (case (proto-set-type field)
+               (case type
                  ((proto:byte-vector cl:string) (= (length value) 0))
-                 ((cl:double-float cl:float) (= value (get-default-form (proto-set-type field)
-                                                                        (proto-default field))))
-                 (t (eq value (get-default-form (proto-set-type field) (proto-default field))))))
+                 ((cl:double-float cl:float) (= value 0))
+                 (t (eq value (eval (proto-initform field))))))
```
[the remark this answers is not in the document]
Slids: Solved by https://github.com/qitab/cl-protobufs/pull/157

serialize.lisp
```diff
@@ -776,112 +788,116 @@ Parameters:
...
         (iterator (if vector-p 'dovector 'dolist))
-        (msg (and class (not (scalarp class))
-                  (or (find-message class)
-                      (find-enum class)
-                      (find-map-descriptor class)))))
-    (cond ((and packed-p (packed-type-p class))
-           `(iincf ,size (serialize-packed ,reader ,class ,index ,vbuf ,vector-p)))
-          ((scalarp class)
-           (let ((tag (make-tag class index)))
+        (desc))
```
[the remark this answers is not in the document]
Slids: Again, what is desc?
Do I set at a desc, is it a descriptor?
A field-descriptor or file-descriptor?

tests/zigzag-test.lisp
```diff
@@ -44,21 +23,14 @@
 (defun expect-bytes (list array)
   (assert-true (equal (coerce list 'list) (coerce array 'list))))
 
-(eval-when (:compile-toplevel :load-toplevel :execute)
-  (proto:define-schema 'zigzag-test
-    :package 'proto-test)
+(defconstant +TAG-S+ (proto-impl::make-tag 'proto:int32 1))
+(defconstant +TAG-U+ (proto-impl::make-tag 'proto:int32 2))
+(defconstant +TAG-I+ (proto-impl::make-tag 'proto:int32 3))
```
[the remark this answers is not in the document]
bkuehnert: Sorry, my commits for this PR were totally out of order. I think this entire PR can be one commit, since every change is a direct result of the changes to model-classes.lisp. 

Sure, I'll make those changes to the tag names.

serialize.lisp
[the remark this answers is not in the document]
bkuehnert: Whoops, thanks for catching that.  #157 should be merged first and this problem disappears.

model-classes.lisp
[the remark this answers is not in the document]
bkuehnert: It is useless at runtime (after #157), that's correct. There is a reason that it's being stored at runtime, and you are right that it's due to problems with the definer macros.

The reason is that we have two separate data structures that store field data, and both are used for generating code. The first data structure is the field-data struct. This is used to define the slots on the protobuf message struct. It needs the field's initform so that the slot has the proper default. The second data structure is the field-descriptor class. This is used for defining the externally facing accessors. This class also needs the initform in order for the clear-* function to work properly.

Once #157 is merged, then there will be no reason to look at `proto-initform` at runtime, so I can delete the slot and just re-compute the initform in the two places that I need it. 

p.s: I also want to remove the `field-data` struct so that this sort of problem never occurs again. Maybe that's a problem for after the internship

serialize.lisp
bkuehnert: Yes, this is OK. If `v` is nil, then there won't be anything to serialize. Would it be more transparent if `boundp` was set to `t`? That should have the same effect. 

With maps, the working assumption is that the type-checked accessors were used to populate the map. So, if we reach this point in the code, `v` MUST be bound to a valid value of type `val-type`, so we should always attempt to serialize it.

protoc/field.cc
```diff
@@ -87,26 +87,73 @@ const std::string FieldLispType(const FieldDescriptor* field) {
...
-      switch (field->options().GetExtension(lisp_container)) {
-        case LIST:
-          return StrCat("(proto:list-of ", type, ")");
-        case VECTOR:
-          return StrCat("(proto:vector-of ", type, ")");
-      }
-    } else {
-      return StrCat("(proto:list-of ", type, ")");
-    }
+const std::string FieldLispClass(const FieldDescriptor* field) {
```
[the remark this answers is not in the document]
bkuehnert: I like category. I'll begin switching to that.

model-classes.lisp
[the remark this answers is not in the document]
bkuehnert: At the moment yeah, there aren't enough uses to warrant adding the new deftypes.

model-classes.lisp
```diff
@@ -459,24 +457,18 @@ on the symbol if we are not in SBCL."
...
-(defconstant $empty-default 'empty-default
-  "The marker used in 'proto-default' used to indicate that there is no default value.")
-(defconstant $empty-list    'empty-list)
-(defconstant $empty-vector  'empty-vector)
-
 ;; Describes a field within a message.
 ;;--- Support the 'deprecated' option (have serialization ignore such fields?)
 (defclass field-descriptor (descriptor)
-  ((type :type string                           ; The name of the Protobuf type for the field
+  ;; The protobuf type of the field. This slot is nullable to allow CCL compilation.
```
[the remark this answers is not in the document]
bkuehnert: In macroexpansion, we get an error like "derived type of :TYPE is (OR NULL SYMBOL) which conflicts with the asserted type of SYMBOL". It happens when we bind :type to something that we got from macroexpanding or destructuring-bind'ing. For example, here: https://github.com/bkuehnert/cl-protobufs/blob/eefeb8273aa1df02f45c33342b96d7f4c3e3d788/define-proto.lisp#L483. 

I tried adding a (declare (type (member ...) type)), which didn't work. Although it's totally possible I did something wrong there.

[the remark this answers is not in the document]
cgay: ok, I wouldn't worry about it then.

model-classes.lisp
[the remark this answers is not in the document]
cgay: ok, let's kick this can down the road some more.  (Not meant to sound accusatory...just frustrating trying to get these macros into decent shape.)

serialize.lisp
cgay: ok, thanks for explaining.  Yes, I think using `t` would be clearer.

protoc/field.cc
```diff
@@ -87,26 +87,73 @@ const std::string FieldLispType(const FieldDescriptor* field) {
...
+  std::string proto_class;
+
+  switch (field->type()) {
+    case FieldDescriptor::TYPE_DOUBLE:
+      proto_class = ":scalar";
+      break;
```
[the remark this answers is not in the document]
bkuehnert: I'd like to avoid fail-throughs if possible

bkuehnert:
> I have to say, this PR is too big. Please try and make them a bit smaller. In this case, for example, the slot renamings could have been in their own PR.

By renaming, you mean the edits I did to the :default slot and :initform slot? Those changes were a direct result to the removal of `proto-set-type`. Additionally, leaving `proto-set-type` alone for the PR wasn't an option either. I don't remember the exact problem, but due to the protoc changes, I think I would've needed to actually add code to keep `proto-set-type` to be the same thing (and at that point, it would just carry a near-duplicate of the :type slot).

protoc/field.cc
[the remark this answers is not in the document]
bkuehnert: On second thought, I'll go with "kind" instead. Category is going to end up being too long.

cgay:
```
Weird number of spaces at line-start.  Are you using a 2-space indent?  [whitespace/indent] [3]
	third_party/lisp/cl_protobufs/protoc/field.cc:139
Warning: messages should be named using CamelCase. If this is a legacy file, consider adding "// LINT: LEGACY_NAMES" to the top of the file. [naming_conventions]
	third_party/lisp/cl_protobufs/tests/zigzag-proto.proto:5:8
```

Slids:
; file: third_party/lisp/cl_protobufs/define-proto.lisp
; in: DEFMACRO DEFINE-MAP
;     (DEFMACRO CL-PROTOBUFS:DEFINE-MAP
;               (PROTOBUFS-IMPLEMENTATION::TYPE-NAME
;                &KEY PROTOBUFS-IMPLEMENTATION::KEY-TYPE
;                PROTOBUFS-IMPLEMENTATION::VAL-TYPE
;                PROTOBUFS-IMPLEMENTATION::VAL-KIND
;                PROTOBUFS-IMPLEMENTATION::INDEX)
;       #<(SIMPLE-BASE-STRING
;          290) Define a lisp type given the data for a protobuf map type.
;   
;   Parameters:
;     TYPE-NAME: Map type name.
;     KEY-TYPE: The lisp type of the map's keys.
;     VAL-TYPE: The lisp type of the map's values.
;     VAL-KI... {100B3EEC0F}>
;       (CHECK-TYPE PROTOBUFS-IMPLEMENTATION::INDEX INTEGER)
;       (LET* ((PROTOBUFS-IMPLEMENTATION::SLOT PROTOBUFS-IMPLEMENTATION::TYPE-NAME)
;              (PROTOBUFS-IMPLEMENTATION::NAME
;               (PROTOBUFS-IMPLEMENTATION::CLASS-NAME->PROTO
;                PROTOBUFS-IMPLEMENTATION::TYPE-NAME))
;              (PROTOBUFS-IMPLEMENTATION::READER
;               (LET #
;                 #))
;              (PROTOBUFS-IMPLEMENTATION::INTERNAL-SLOT-NAME
;               (PROTOBUFS-IMPLEMENTATION::FINTERN "%~A"
;                                                  PROTOBUFS-IMPLEMENTATION::SLOT))
;              (PROTOBUFS-IMPLEMENTATION::QUAL-NAME
;               (PROTOBUFS-IMPLEMENTATION::MAKE-QUALIFIED-NAME
;                PROTOBUFS-IMPLEMENTATION::*CURRENT-MESSAGE-DESCRIPTOR* #))
;              (CLASS (PROTOBUFS-IMPLEMENTATION::FINTERN #))
;              (PROTOBUFS-IMPLEMENTATION::MSLOT
;               (PROTOBUFS-IMPLEMENTATION::MAKE-FIELD-DATA :INTERNAL-SLOT-NAME
;                PROTOBUFS-IMPLEMENTATION::INTERNAL-SLOT-NAME :EXTERNAL-SLOT-NAME
;                PROTOBUFS-IMPLEMENTATION::SLOT :TYPE 'HASH-TABLE :INITFORM '#
;                :ACCESSOR PROTOBUFS-IMPLEMENTATION::READER))
;              (PROTOBUFS-IMPLEMENTATION::MFIELD
;               (MAKE-INSTANCE 'CL-PROTOBUFS:FIELD-DESCRIPTOR :NAME # :TYPE CLASS
;                              :KIND :MAP :QUALIFIED-NAME
;                              PROTOBUFS-IMPLEMENTATION::QUAL-NAME :LABEL :OPTIONAL
;                              ...))
;              (PROTOBUFS-IMPLEMENTATION::MAP-DESC
;               (PROTOBUFS-IMPLEMENTATION::MAKE-MAP-DESCRIPTOR :NAME
;                                                              PROTOBUFS-IMPLEMENTATION::NAME
;                                                              :KEY-TYPE
;                                                              PROTOBUFS-IMPLEMENTATION::KEY-TYPE
;                                                              :VAL-TYPE
;                                                              PROTOBUFS-IMPLEMENTATION::VAL-TYPE
;                                                              :VAL-KIND
;                                                              PROTOBUFS-IMPLEMENTATION::VAL-KIND)))
;         (PROTOBUFS-IMPLEMENTATION::RECORD-PROTOBUF-OBJECT CLASS
;                                                           PROTOBUFS-IMPLEMENTATION::MAP-DESC
;                                                           :MAP)
;         `(PROGN
;           CL-PROTOBUFS:DEFINE-MAP
;           PROTOBUFS-IMPLEMENTATION::MAP-DESC
;           ((PROTOBUFS-IMPLEMENTATION::RECORD-PROTOBUF-OBJECT ',CLASS
;                                                              ,PROTOBUFS-IMPLEMENTATION::MAP-DESC
;                                                              :MAP))
;           ,PROTOBUFS-IMPLEMENTATION::MFIELD
;           ,PROTOBUFS-IMPLEMENTATION::MSLOT)))
; --> SB-INT:NAMED-DS-BIND SB-INT:BINDING* LET* IF 
; ==>
;   NIL
; 
; caught STYLE-WARNING:
;   This is not a (MEMBER :SCALAR :MESSAGE :GROUP :ENUM):
;    NIL
;   See also:
;     The SBCL Manual, Node "Handling of Types"
ERROR: FAIL: SB-INT:TYPE-STYLE-WARNING 'This is not a 
         (MEMBER :SCALAR :MESSAGE :GROUP :ENUM):
 NIL
See also:
  The SBCL Manual, Node "Handling of Types"'

Slids:
Where are we now:

Done:
LISP-TYPE has been removed. It didn't do anything with the removal of aliased types.
CONTAINER has been added as a slot that stores either :LIST, :VECTOR, or NIL. This indicates which type of lisp container is used for a repeated field (or NIL if the field is not repeated). This information is required by the deserializer, and was previously pulled out of the DEFAULT slot.

I think done:
(Needs to be renamed initform)
DEFAULT has been removed and replaced with INITFORM. Previously default held either a default or a symbol which indicated if the default was an empty list or vector. This was confusing and unnecessary, and it required re-computing a type's default several times.


TODO

The TYPE slot has been repurposed. Instead of storing a string that named the protobuf type, it now stores the base lisp type for the field. "Base" here means that if a type is repeated, then TYPE is bound to [type] rather than (list-of [type]). This is because this information is duplicated by the label and container fields anyway.
The CLASS slot has been repurposed. Instead of storing the lisp type (if it is an enum, map, or message) or a keyword (if it's scalar), it now stores a keyword which indicates if the type is an enum, map, message, or scalar. This is useful since we often need to distinguish between these different classes of types, and previously this was impossible at compile time.
SET-TYPE has been removed. It previously stored the lisp type.

PR #157, opened 2020-08-11
Use a plist on the symbol instead of hashmap for proto field accessors

Slids:
It's cheaper and easier to save the accessor, getter, and has functions
on the plist then to define a global hash-map like we curently do.
So use the plist version.
The performance difference seems to be a wash.

### DIFF
--- a/define-proto.lisp
+++ b/define-proto.lisp
@@ -520,21 +520,23 @@ Parameters:
   "Sets the %bytes field of the proto object OBJ with NEW-VALUE."
   (setf (slot-value obj '%bytes) new-value))
 
-(defstruct field-values
+(defstruct field-accessors
   "Structure containing the get, set, and has functions
 for a proto-message field."
   (get nil :type symbol)
   (set nil :type list)
   (has nil :type symbol))
 
-(defun set-field-value-functions (proto-type field-name)
-  "Set the get, set, and has functions for a field onto
-the symbol plist inside of a field-values struct."
-  (setf (get field-name proto-type)
-        (make-field-values
-         :get (proto-slot-function-name proto-type field-name :get)
-         :set `(setf ,(proto-slot-function-name proto-type field-name :get))
-         :has (proto-slot-function-name proto-type field-name :has))))
+(defun set-field-value-functions (message-name field-symbol)
+  "Set the get, set, and has functions for a proto field on a fields symbol p-list.
+Parameters:
+  MESSAGE-NAME: The name of the protobuf message containing the field.
+  FIELD-SYMBOL: The symbol for the field."
+  (setf (get field-symbol message-name)
+        (make-field-accessors
+         :get (proto-slot-function-name message-name field-symbol :get)
+         :set `(setf ,(proto-slot-function-name message-name field-symbol :get))
+         :has (proto-slot-function-name message-name field-symbol :has))))
 
 (defun make-common-forms-for-structure-class (proto-type public-slot-name slot-name field)
   "Create the common forms needed for all message fields
--- a/define-proto.lisp
+++ b/define-proto.lisp
@@ -525,7 +525,8 @@ Parameters:
 for a proto-message field."
   (get nil :type symbol)
   (set nil :type list)
-  (has nil :type symbol))
+  (has nil :type symbol)
+  (clear nil :type symbol))
 
 (defun set-field-accessor-functions (message-name field-name)
   "Set the get, set, and has functions for a proto field on a fields symbol p-list.
@@ -536,7 +537,8 @@ Parameters:
         (make-field-accessors
          :get (proto-slot-function-name message-name field-name :get)
          :set `(setf ,(proto-slot-function-name message-name field-name :get))
-         :has (proto-slot-function-name message-name field-name :has))))
+         :has (proto-slot-function-name message-name field-name :has)
+         :clear (proto-slot-function-name message-name field-name :clear))))
 
 (defun make-common-forms-for-structure-class (proto-type public-slot-name slot-name field)
   "Create the common forms needed for all message fields
--- a/define-proto.lisp
+++ b/define-proto.lisp
@@ -174,10 +174,11 @@ the oneof and its nested fields.
   "Validates that all of the IMPORTS (a list of file names) have
    already been loaded. FILE-DESCRIPTOR is the descriptor of the
    file doing the importing."
-  (dolist (import (reverse imports))
-    (let* ((imported (proto:find-schema (if (stringp import) (pathname import) import))))
-      (unless imported
-        (error "Could not find file ~S imported by ~S" import file-descriptor)))))
+  ;; (dolist (import (reverse imports))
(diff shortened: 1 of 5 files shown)


serialize.lisp
```diff
@@ -131,43 +131,34 @@ Parameters:
...
   BUFFER: The buffer to serialize to."
   (declare (type field-descriptor field))
-  (flet ((read-slot (object slot reader)
-           ;; Don't do a boundp check, we assume the object is fully populated.
-           ;; Unpopulated slots should be "nullable" and will contain nil when empty.
-           (if reader
-               (funcall reader object)
-               (slot-value object slot))))
+  (unless
+      (if (eq (slot-value field 'message-type) :extends)
```
bkuehnert, line 135: Why don't we call the has-* function from the symbol p-list at this point?

serialize.lisp
```diff
@@ -131,43 +131,34 @@ Parameters:
...
+                       (t (proto-slot-value object (slot-value field 'external-field-name))))))
+    ;; For singular fields, only emit if VALUE is not default.
     (when
-        (or (and (eq (slot-value field 'message-type) :extends)
-                 (not (has-extension object (slot-value field 'internal-field-name))))
-            (and (slot-value field 'field-offset)
-                 (= (bit (slot-value object '%%is-set)
-                    (slot-value field 'field-offset))
-                    0)))
+        (and (eq (proto-label field) :singular)
```
bkuehnert, line 148: Same thing here. If the has-* function is called above, then this `when` is not necessary.

bkuehnert:
text-format.lisp should also be updated

Slids:
This is my code and I'm a googler.

text-format.lisp
```diff
@@ -30,68 +30,54 @@ Parameters:
...
+          (format stream "{")))
+    (dolist (field (proto-fields message))
+      (when (if (eq (slot-value field 'message-type) :extends)
+                (has-extension object (slot-value field 'internal-field-name))
+                (has-field object (slot-value field 'external-field-name)))
+        (let* ((value
+                 (cond ((eq (slot-value field 'message-type) :extends)
+                        (get-extension object (slot-value field 'external-field-name)))
+                       ((proto-lazy-p field)
+                        (slot-value object (slot-value field 'internal-field-name)))
```
bkuehnert, line 47: We don't actually want to call slot-value here. If this field is lazy deserialized, then calling slot-value will return an empty message. We want to use the reader in this case, as this will deserialize the lazy bytes. Simply removing this case should fix this.

text-format.lisp
```diff
@@ -30,68 +30,54 @@ Parameters:
...
-              (if (eq (proto-label field) :repeated)
-                  (print-repeated-field value
+    (let ((name (or name (proto-name message))))
+      (if print-name
+          (if pretty-print
+              (format stream "~&~V,0T~A {~%" indent name)
+              (format stream "~A { " name))
+          (format stream "{")))
+    (dolist (field (proto-fields message))
+      (when (if (eq (slot-value field 'message-type) :extends)
```
cgay, line 40: `print-text-format` doesn't seem performance critical, so use the field accessor instead of `slot-value`?  Same below.

define-proto.lisp
```diff
@@ -520,24 +520,21 @@ Parameters:
...
-    (setf (gethash field-name value)
-          (list (proto-slot-function-name proto-type field-name :has)
-                (proto-slot-function-name proto-type field-name :get)
-                (proto-slot-function-name proto-type field-name :clear)))))
+(defstruct field-values
+  "Structure containing the get, set, and has functions
+for a proto-message field."
+  (get nil :type symbol)
+  (set nil :type list)
+  (has nil :type symbol))
```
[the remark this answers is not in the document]
Slids: Your comment:
field-accessors-{get,set,has} sound sgood.
I don't haser, just sounds dumb...

text-format.lisp
Slids: You need to be more exact with which slot-value you mean...

serialize.lisp
```diff
@@ -131,43 +131,25 @@ Parameters:
...
-              (undefined-field-type "While serializing ~S,"
-                                    object type field))))))
+  (unless
+      (if (eq (slot-value field 'message-type) :extends)
+          (has-extension object (slot-value field 'internal-field-name))
+          (has-field object (slot-value field 'external-field-name)))
+    (return-from emit-field 0))
+  (let* ((type   (slot-value field 'class))
+         (index  (proto-index field))
+         (value  (cond ((eq (slot-value field 'message-type) :extends)
```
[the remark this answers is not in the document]
Slids: I think this is okay...
It seems not important to save the value when it only
might be called at most twice, and it's not like the slot-value
call here is expensive.
If it were used more, or in other paths I'd agree, but...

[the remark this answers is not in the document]
Slids: ah sure...

text-format.lisp
Slids: Why?
It's a known good function in common lisp thats
gauranteed to work across message types.
The generic function accessor doesn't provide any more
clarity then the slot value call...

define-proto.lisp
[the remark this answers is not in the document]
cgay: What I would do here is to use `conc-name: ""` and make these be the slot names:
```
field-getter
field-setter
field-is-set-function
```

text-format.lisp
cgay: The main reason has nothing to do with clarity, it's about being able to find the xrefs. We've talked about this.

Whatever...the horse has already left the barn.

define-proto.lisp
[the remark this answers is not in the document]
Slids: I like that even less...
You should leave the conc name so it is dramatically  apparent
your referencing into a struct.

define-proto.lisp
```diff
@@ -520,24 +520,23 @@ Parameters:
...
-                (proto-slot-function-name proto-type field-name :get)
-                (proto-slot-function-name proto-type field-name :clear)))))
+(defstruct field-accessors
+  "Structure containing the get, set, and has functions
+for a proto-message field."
+  (get nil :type symbol)
+  (set nil :type list)
+  (has nil :type symbol))
+
+(defun set-field-value-functions (message-name field-symbol)
```
cgay, line 530: This should probably match the struct name: `set-field-accessors`? `set-field-accessor-functions`?

define-proto.lisp
```diff
@@ -520,24 +520,23 @@ Parameters:
...
+  "Set the get, set, and has functions for a proto field on a fields symbol p-list.
+Parameters:
+  MESSAGE-NAME: The name of the protobuf message containing the field.
```
cgay, line 533: Is this a string? The juxtaposition with the other parameter being named `-symbol` makes me think so.

Slids: No, it's not a string.

Slids:
MErged as https://github.com/qitab/cl-protobufs/pull/160

PR #158, opened 2020-08-11
Fix text format lazy bug

bkuehnert:
This code was lazily copied from the serializer code, in which we want
to hold off from deserializing lazy fields when serializing. Since we
are printing to text, we actually need to deserialize. So, this check
for lazy fields is bad.


bkuehnert:
Closing. #157 fixes the bug

PR #159, opened 2020-08-12
Add a benchmarks test suite

Slids:
Currently we do some really simple benchmarking in serialize and proto3
We should move them to their own test suite and do some tests of  basic functionality.
Not perfect but a first step.


tests/timing-tests.lisp
```diff
@@ -0,0 +1,199 @@
...
+        (setf (proto3-proto:person.id spouse) 123456)
+        (setf (proto3-proto:person.name spouse) "Spouse")
+        (setf (proto3-proto:person.id person) count)
+        (setf (proto3-proto:person.name person) name)
+        (setf (proto3-proto:person.home person) address)
+        (setf (proto3-proto:person.spouse person) spouse)
+        (setf (proto3-proto:person.odd-p person) (oddp count))
+        (push person (proto3-proto:population.people population))))
+    (print "Proto3 Performance Test Serialization")
+    (benchmark:with-timing (1000)
```
cgay, line 122: Seems like `with-timing` should have a way to name the code being timed so you don't have to `print` in front of it...

Slids: I wouldn't really want it to,
Just a random but of code out of the context is kinda messy 
and wastes space...

tests/timing-tests.lisp
```diff
@@ -0,0 +1,199 @@
...
+        #:cl-protobufs)
+  (:local-nicknames
+   (#:serial-proto #:cl-protobufs.serialization-test)
+   (#:oneof-proto #:cl-protobufs.oneof-test)
+   (#:proto3-proto #:cl-protobufs.proto3-test)
+   (#:map-proto #:cl-protobufs.map-test)))
+
+(in-package #:cl-protobufs.test.timing-tests)
+
+(defsuite timing-tests (cl-protobufs.test:root-suite))
```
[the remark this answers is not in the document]
Slids: /sigh, meh...

PR #161, opened 2020-08-13
Fix maps using incorrect :test function.

bkuehnert:
Now, the underlying lisp hash-table will use 'eq' if the key type is
integral, or 'equal' if the key type is string. This is more efficient
in the integral case, and works properly in the string case (eql doesn't
work for string equality).

Due to some weirdness in the definer macros, this requires the clear-*
function to be defined in the make-map-accessors defun. Admittedly, this
is also due to me doing a poor job when implementing map types, but I
plan to fix this.

### DIFF
--- a/define-proto.lisp
+++ b/define-proto.lisp
@@ -596,15 +596,16 @@ Arguments:
 
         ;; Clear function
         ;; Map type clear functions are created in make-map-accessor-forms.
-        ,(unless (eq (proto-set-type field) :map)
-           `(declaim (inline ,clear-function-name))
-           `(defun ,clear-function-name (,obj)
-              ,(when index
-                 `(setf (bit (,is-set-accessor ,obj) ,index) 0))
-              ,(if bool-index
-                   `(setf (bit (,bit-field-name ,obj) ,bool-index)
-                          ,(if default-form 1 0))
-                   `(setf (,hidden-accessor-name ,obj) ,default-form))))
+        ;; todo(benkuehnert): rewrite map types/definers so that this isn't necessary
+        ,@(unless (eq (proto-set-type field) :map)
+            `((declaim (inline ,clear-function-name))
+              (defun ,clear-function-name (,obj)
+                ,(when index
+                   `(setf (bit (,is-set-accessor ,obj) ,index) 0))
+                ,(if bool-index
+                     `(setf (bit (,bit-field-name ,obj) ,bool-index)
+                            ,(if default-form 1 0))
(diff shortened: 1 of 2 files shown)


define-proto.lisp
```diff
@@ -593,14 +595,16 @@ Arguments:
...
-        (declaim (inline ,clear-function-name))
-        (defun ,clear-function-name (,obj)
-          ,(if bool-index
-               `(setf (bit (,bit-field-name ,obj) ,bool-index)
-                      ,(if default-form 1 0))
-               `(setf (,hidden-accessor-name ,obj) ,default-form))
-          ,(when index
-            `(setf (bit (,is-set-accessor ,obj) ,index) 0)))
+        ;; Map type clear functions are created in make-map-accessor-forms.
+        ,(unless (eq (proto-set-type field) :map)
```
Slids, line 599: add a comment above 
;; todo(bkuehnert): Fix map type so we can craete the clear functions here.

tests/map-test.lisp
```diff
@@ -62,6 +62,15 @@
     (assert-equal (map-all.msgmap-gethash 0 test) nil)
     (assert-equal (map-all.enummap-gethash 0 test) :one)))
 
+;; Verify that the lisp hash-table properly handles string keys
+(deftest string-key-check (map-suite)
```
Slids, line 66: Add a mixed case test, like
"Foo"
and "foo"

define-proto.lisp
bkuehnert: Fixing this also requires removing the field-data struct, so it's not *all* my fault. Adding a comment anyway.

define-proto.lisp
```diff
@@ -593,14 +595,17 @@ Arguments:
...
-          ,(if bool-index
-               `(setf (bit (,bit-field-name ,obj) ,bool-index)
-                      ,(if default-form 1 0))
-               `(setf (,hidden-accessor-name ,obj) ,default-form))
-          ,(when index
-            `(setf (bit (,is-set-accessor ,obj) ,index) 0)))
+        ;; Map type clear functions are created in make-map-accessor-forms.
+        ;; todo(benkuehnert): rewrite map types/definers so that this isn't necessary
+        ,(unless (eq (proto-set-type field) :map)
+           `(declaim (inline ,clear-function-name))
```
cgay, line 601: This declaim form is dropped on the floor. Only the value of the last form in the unless becomes part of the returned list.

PR #164, opened 2020-08-15
Save proto bytes that we can't deserialize for later

Slids:
Suppose we have a proto message foo at version 1 and 2:
Version 1:
message foo {
  optional int one = 1;
}
Version 2:
message foo {
  optional int one = 1;
  optional int two = 2;
}

If we serialize a message at version 2, and deserialize the
message at version 1, we don't necessarily want to drop the bytes
If we serialize the the deserialized foo we should attach the 
skipped bytes to the new byte array so a future recipient can 
read them properly.

### DIFF
--- a/serialize.lisp
+++ b/serialize.lisp
@@ -129,12 +129,13 @@
 (defun emit-skipped-bytes (object buffer)
   (declare (buffer buffer)
            (base-message object))
-  (let ((skipped-byte-length (length (base-message-%%skipped-bytes object))))
-    (if (zerop skipped-byte-length)
-        0
-        (progn
-          (buffer-ensure-space buffer skipped-byte-length)
-          (fast-octets-out buffer (base-message-%%skipped-bytes object))))))
+  (if (base-message-%%skipped-bytes object)
+      (progn
+        (buffer-ensure-space
+         buffer
+         (length (base-message-%%skipped-bytes object)))
+        (fast-octets-out buffer (base-message-%%skipped-bytes object)))
+      0))
 
 (defun emit-field (object field buffer)
   "Serialize a single field from an object to buffer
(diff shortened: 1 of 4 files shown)


Slids:
This is part of the documentaion:
https://developers.google.com/protocol-buffers/docs/overview#updating

serialize.lisp
```diff
@@ -121,7 +121,21 @@
...
-                   (emit-non-repeated-field value type index buffer))))))))
+                   (emit-non-repeated-field value type index buffer))))))
+
+    (incf size (emit-skipped-bytes object buffer))))
+
+
+(defun emit-skipped-bytes (object buffer)
+  (declare (buffer buffer)
+           (base-message object))
+  (if (base-message-%%skipped-bytes object)
```
bkuehnert, line 132: if-bind here so that we don't have to repeatedly call `(base-message-%%skipped-bytes object)` ?

tests/serialization-tests.lisp
```diff
@@ -549,16 +549,29 @@
...
+           (proto-different-than-wire (make-proto-different-than-wire
+                                       :beginning "char"
+                                       :always "pika-pal")))
+      (assert-true my-deserialized-proto)
+      (assert-true (proto-equal my-deserialized-proto proto-different-than-wire))
+      (let* ((reserializaed-proto-octets (serialize-object-to-bytes my-deserialized-proto))
+             (should-be-original-proto
+               (deserialize-object 'proto-on-wire
+                                   reserializaed-proto-octets)))
+        (assert-true (proto-equal should-be-original-proto proto-on-wire))))))
```
cgay, line 577: This test will only run correctly the first time and thereafter it will not be testing the unoptimized serializers.

PR #173, opened 2020-08-21
Add more general handling for 'null' fields.

bkuehnert:
I missed one piece of syntax in the original PR. This PR fixes that. 

'null' can also appear as the value for a non-repeated field. This value
represents the default value of the type. Since JSON mappings are only
documented for proto3 semantics, this means that we should simply not
set the field to anything in this case. In proto2, this means that the
has-* function should return NIL if a field is parsed with a 'null'
value. This commit adds a test for this behavior.

### DIFF
--- a/json.lisp
+++ b/json.lisp
@@ -200,31 +200,36 @@ Parameters:
              (slot  (and field (proto-external-field-name field))))
         (expect-char stream #\:)
         (if (null field)
+            ;; If FIELD is null, then we assume that MSG-DESC describes a
+            ;; different version of the proto on the wire which doesn't
+            ;; have FIELD, and continue,
+            ;; todo(benkuehnert): Add a flag to optionally throw an error
+            ;; in this spot.
             (skip-json-value stream)
-            (let (val error-p)
-              (if (not (eq (proto-label field) :repeated))
-                  (multiple-value-setq (val error-p)
-                    (parse-value-from-json type :stream stream))
-                  (case (peek-char nil stream nil)
-                    ;; Repeated field is in list format
-                    ((#\[)
-                     (expect-char stream #\[)
-                     (loop
-                       (multiple-value-bind (data err)
-                           (parse-value-from-json type :stream stream)
-                         (if err
-                             (setf error-p t)
-                             (push data val)))
-                       (if (eql (peek-char nil stream nil) #\,)
-                           (expect-char stream #\,)
-                           (return)))
-                     (expect-char stream #\]))
-                    ;; 'null' is parsed as an empty list. Anything else is an error.
-                    ((#\n)
-                     (let ((tok (parse-token stream)))
-                       (unless (string= tok "null")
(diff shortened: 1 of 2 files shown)


bkuehnert:
> 
> 
> > Locally, ABCL tests are failing due to a type error in full-tests.lisp. This code has absolutely nothing to do with the change, so I think this might be due to something being wrong with my local install. Not sure what is causing the failure in CI.
> 
> Nope, failing on github ci as well

I know. The failure I'm seeing locally is not the one that's causing it to fail on CI. I have not yet been able to reproduce the error message that CI reports.

Slids:
Note: It's definitely not this pr as we have the same failure at head...

bkuehnert:
> Note: It's definitely not this pr as we have the same failure at head...

Yes, it is something to do with the removal of `fold-symbol`, and something about inlined functions. 

The failure I was seeing locally was because of an outdated protoc plugin.

json.lisp
```diff
@@ -200,31 +200,33 @@ Parameters:
...
-                     (expect-char stream #\]))
-                    ;; 'null' is parsed as an empty list. Anything else is an error.
-                    ((#\n)
-                     (let ((tok (parse-token stream)))
-                       (unless (string= tok "null")
-                         (setf error-p t))))))
+            (let (val error-p null-p)
               (cond
+                ;; If we see an 'n', then the value MUST be 'null'. I
+                ;; this case, parse the 'null' and continune.
```
cgay, line 208: "I" should be "In" and continue is misspelled.

json.lisp
```diff
@@ -200,31 +200,33 @@ Parameters:
...
+                ((eql (peek-char nil stream nil) #\n)
+                 (parse-token stream)
+                 (skip-whitespace stream)
+                 (setf null-p t))
```
cgay, line 212: You should verify that the parsed token is in fact "null".

json.lisp
```diff
@@ -200,31 +200,33 @@ Parameters:
...
+                         (setf error-p t)
+                         (push data val)))
+                   (if (eql (peek-char nil stream nil) #\,)
+                       (expect-char stream #\,)
+                       (return)))
+                 (expect-char stream #\]))
+                (t (multiple-value-setq (val error-p)
+                    (parse-value-from-json type :stream stream))))
+              (cond
+                ;; If we read a null, do nothing.
```
cgay, line 228: This comment doesn't really add anything; the code is simple and clear already.

tests/json-test.lisp
```diff
@@ -50,7 +50,8 @@
   \"int64_field\": \"12\",
   \"sint64_field\": \"-1\",
   \"int_vals\": null,
-  \"oneof_int_field\": 5
+  \"oneof_int_field\": 5,
+  \"embedded_comment_a\": null
```
cgay, line 54: Is there a field in this json that doesn't exist in the .proto file, to test for skipping missing fields?

PR #177, opened 2020-08-24
JSON Well known types

bkuehnert:
This PR adds support for well known types' special JSON mappings.


Slids:
Interesting, abcl gives: 
; Caught STYLE-WARNING:                                                                                                                                
;   Undefined variable D0 assumed special                                                                                                              
                                                                                                                                                       
; Caught COMPILE-WARNED-WARNING:                                                                                                                       
;   Lisp compilation had style-warnings while compiling #<ASDF/LISP-ACTION:CL-SOURCE-FILE "cl-protobufs" "parsing" "parser">

json.lisp
```diff
@@ -200,31 +213,33 @@ Parameters:
              (slot  (and field (proto-external-field-name field))))
         (expect-char stream #\:)
         (if (null field)
+            ;; Should we signal an error here?
```
Slids, line 216: When I find myself in times of trouble,
I always ask,
What would c++ do?
Okay, no I don't but what would it do?

json.lisp
```diff
@@ -348,3 +363,177 @@ be either an array, object, or primitive."
            thereis (find name (oneof-descriptor-fields oneof)
                          :key #'proto-json-name
                          :test #'string=))))
+
+;; Special JSON mappings for well known types below
+
+(defun special-json-p (type)
+  "Check if the message TYPE has a special JSON mapping."
+  (member type '(cl-protobufs.google.protobuf:any
```
Slids, line 371: use a local-nickname,
this verbosity is silly...

json.lisp
```diff
@@ -348,3 +363,177 @@ be either an array, object, or primitive."
...
+                 cl-protobufs.google.protobuf:float-value
+                 cl-protobufs.google.protobuf:int32-value
+                 cl-protobufs.google.protobuf:int64-value
+                 cl-protobufs.google.protobuf:u-int32-value
+                 cl-protobufs.google.protobuf:u-int64-value)))
+
+(defun wrapper-message->type (type)
+  "For a well known wrapper type TYPE, return the type being wrapped."
+  (ecase type
+    ((cl-protobufs.google.protobuf:bool-value) :bool)
```
Slids, line 391: whats the difference between these and those in
special-json-p?
Seem to be less here...
Why?

Slids: Also, local-nickname

parser.lisp
```diff
@@ -232,20 +232,25 @@
     (when number
       (coerce number 'float))))
 
-(defun parse-double (stream)
+(defun parse-double (stream &key append-d0)
   "Parse the next token in the STREAM as a double, then skip the following whitespace.
-   The returned value is the double-float."
-  (let ((number (parse-number stream)))
+If APPEND-D0 is true, then append "d0" to the parsed number before attempting to convert
+to a double. The returned value is the double-float."
```
Slids, line 238: Why?
Seems strange, please add detail as to why...

tests/json-test.lisp
```diff
@@ -98,16 +109,109 @@
...
+    (json-roundtrip int32-wrap)
+    (json-roundtrip int64-wrap)
+    (json-roundtrip uint32-wrap)
+    (json-roundtrip uint64-wrap)))
+
+;; This tests the JSON parser/printer handling any messages that contain packed well known types.
+(deftest special-json-nested (json-suite)
+  (let* ((nested (make-text-format-test.nested-message1 :int-field 2))
+         (msg (make-text-format-test :int-field 100 :float-field 1.5 :one-level-nesting nested))
+         (timestamp (cl-protobufs.google.protobuf:make-timestamp :seconds 1598289366 :nanos 4000))
```
Slids, line 166: local-nickname,

well-known-types.lisp
```diff
@@ -43,22 +43,26 @@
...
+            type-part-of-url)))
     (assert type ()
             "Could not find class for type: ~S." type-part-of-url)
+    type))
+
+(defun unpack-any (any-message)
+  "Given an Any message unpack retrieve the stored proto message.
+Parameters:
+  ANY-MESSAGE: The message to unpack."
+  (let ((type (resolve-type-url (cl-protobufs.google.protobuf:any.type-url any-message)))
```
Slids, line 64: local-nickname

json.lisp
bkuehnert: The answer is that C++ code does signal an error in proto3. In proto2, there is no official JSON mapping so there's no answer.

I'll add a separate PR which adds the error.

json.lisp
bkuehnert: From the comment: these are the wrapper types. This function translates between the protobuf type and the underlying `proto-class` data, so that we don't have to repeat code anywhere.

parser.lisp
bkuehnert: Yeah, it is strange and I don't like it at all. The reason is because this code was used to parse strings of the form "3.14d0" instead of "3.14". Turns out, that "d0" is necessary in order for lisp to interpret it as a double-float. 

I'm not very familiar with how lisp handles floating point, but I searched for a way to parse numbers as doubles and there doesn't seem to be a clean way.

Another option would be: Write another float/double parser which just parses the integer part & decimal part separately as numbers, and then combine them numerically. This is distinct from what we do now, which is just `read-from-string` and then type coercion.

json.lisp
cgay: Let's only review this code in a single PR, the other one, ok?

Issue #181, opened 2020-08-25
Map field w/ message type value defined after the map exhibits type error


bkuehnert:
Consider the following proto:
```
syntax = "proto2"

message Foo {
  map<string,Bar> map_field = 1;
}

message Bar {
  optional int32 int_field = 1;
}
```

When the accessor `foo.map-field-gethash` is being generated, cl-protobufs calls `(find-message bar)` to check if `bar` constitutes a message type. This will return `nil` as the message has not yet been processed yet. So, the output type of `foo.map-field-gethash` gets set to `bar`. 

This is a problem, since `nil` is the default for any message type, and `nil` is not of type `bar`. Hence, calling the accessor on any un-set key wil result in a type error.

The simple fix is to just set the underlying type of the hash-table to be `(or null message)` via protoc, which is what we do for all message fields. Sadly, this allows us to insert `nil` into the map, but we cannot get around this until there is a proper `proto-kind` slot on the field descriptor (or outputted by protoc).

Issue #182, opened 2020-08-25
cl-protobufs should import struct.proto


bkuehnert:
[struct.proto](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/struct.proto) is a well known type, and thus should be loaded by the well known types module.

The messages (and one enum!) have special JSON mappings which will also need to be implemented. 

This shouldn't be done until #181 is fixed.

PR #186, opened 2020-08-26
Fix map field type error when value type is defined after the map

bkuehnert:
This fixes bug #181 and adds tests for the bug. With this change, now
NIL can be inserted into maps that have message values. This is a bug,
but it is the same behavior as any optional message field.


Slids:
Please don't make descriptions like: Fix bug #181 #186
I have no idea what bug 181 is, instead say what it's doing and then link to the bug
in the description.

bkuehnert:
> Please don't make titles like: Fix bug #181 #186
> I have no idea what bug 181 is, instead say what it's doing and then link to the bug
> in the description.

Edited it. The bug is a bit of a mouthful, hence why I linked it

Slids:
Should this also update the json code?
Or is that in another cl?

bkuehnert:
> Should this also update the json code?
> Or is that in another cl?

What would change in the JSON code?

Slids:
You commented out several well known types.

Slids:
I see, this is cl-protobufs should import struct.proto #182
nvm

bkuehnert:
> I see, this is cl-protobufs should import struct.proto #182
> nvm
> For some reason I thought this and that were related...

Oh right, I understand. That's next :)

model-classes.lisp
```diff
@@ -308,8 +308,8 @@ Parameters:
   (name      nil :type string)
   (key-class nil :type symbol) ;; the :class of the key
   (val-class nil :type symbol) ;; the :class of the value
-  (key-type nil :type symbol)  ;; the lisp type of the key
-  (val-type nil :type symbol)) ;; the lisp type of the value
+  (key-type nil)  ;; the lisp type of the key, untyped
+  (val-type nil)) ;; the lisp type of the value, untyped.
```
cgay, line 312: I don't think "untyped" adds any information here. Either leave it off or say "e.g., (or null msg)"

tests/map-proto.proto
```diff
@@ -24,13 +24,14 @@ message MapProto {
...
-  message ValMessage {
-    optional string strval = 1;
-  }
   map<int32,ValMessage> map_field = 1;
   optional int32 after = 3;
 }
 
+message ValMessage {
+    optional string strval = 1;
+  }
```
cgay, line 33: indentation

Also please add a comment that this is intentionally defined after its use in MapMessage.

define-proto.lisp
```diff
@@ -448,7 +448,11 @@ Parameters:
                     :class class
                     :name name
                     :key-class (lisp-type-to-protobuf-class key-type)
-                    :val-class (lisp-type-to-protobuf-class val-type)
+                    ;; If the value type is a message, then VAL-TYPE will take the form
+                    ;; (cl:or cl:null message). In this case, the class is just message.
+                    :val-class (if (listp val-type)
```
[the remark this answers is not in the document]
bkuehnert: That's much better, thank you.

As for the comment: This doesn't disallow nil being inserted into the map. This just sets up the `val-class` slot on the map-descriptor so that it can be used with the (de)serialization functions. They expect messages to be given in the form `message` rather than `(or null message)`. I'm updating accordingly

Slids:
merged as of https://github.com/qitab/cl-protobufs/commit/27d33c63b1516662e43e633fad15728d2034bebd

PR #188, opened 2020-08-26
Remove 'singular' label

bkuehnert:
This label is unecessary and confusing as some proto documentation uses
'singular' as a synonym for 'optional'. Since we only need to
distinguish between proto2 and proto3 at compile time, there's no need
to hang on to this information.

### DIFF
--- a/define-proto.lisp
+++ b/define-proto.lisp
@@ -556,12 +556,11 @@ Parameters:
         (index (proto-field-offset field))
         (clear-function-name (proto-slot-function-name proto-type public-slot-name :clear))
         (bool-index (proto-bool-index field))
-        (bit-field-name (fintern "~A-%%BOOL-VALUES" proto-type))
-        (label (proto-label field)))
+        (bit-field-name (fintern "~A-%%BOOL-VALUES" proto-type)))
 
     ;; If index is nil, then this field does not have a reserved bit in the %%is-set vector.
-    ;; This means that the field is singular, so checking for field presence must be done by
-    ;; checking if the bound value is default.
+    ;; This means that the field is proto3-style optional, so checking for field presence must
+    ;; be done by checking if the bound value is default.
     (with-gensyms (obj new-value cur-value)
       `(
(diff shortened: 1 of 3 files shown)


define-proto.lisp
```diff
@@ -613,9 +612,9 @@ Parameters:
 
         (proto-impl::set-field-accessor-functions ',proto-type ',public-slot-name)
 
-        ;; has-* functions are not exported for singular fields. They are only for
+        ;; has-* functions are not exported for proto3-style optional fields. They are only for
         ;; internal usage.
-        ,(unless (eq label :singular)
+        ,(unless (eq (proto-syntax *current-file-descriptor*) :proto3)
```
Slids, line 617: hows does this work for proto2-style optionals in a proto3 file?

bkuehnert: proto2-style optionals in a proto3 file are wrapped inside of oneofs. So, this function never gets called for those fields. Instead, they get handled via the `make-oneof-accessor-forms` function, which doesn't care about labels.

Slids: please add a comment then

define-proto.lisp
```diff
@@ -1587,13 +1586,11 @@ function) then there is no guarantee on the serialize function working properly.
...
-          ;; Proto3 optional fields are known as 'singular' fields, and are handled differently.
-          (let* ((label (if (and (eq (proto-syntax *current-file-descriptor*) :proto3)
-                                 (eq label :optional))
-                            :singular
-                            label))
-                 ;; Singular fields do not have offsets, as they don't have has-* functions.
-                 (offset (and (not (eq label :singular)) field-offset))
+          (let* (;; Proto3 optional fields do not have offsets, as they don't have has-* functions.
+                 ;; Note that proto2-style optional fields in proto3 files are wrapped in oneofs by
+                 ;; protoc, and hence process-field never get called for them.
```
cgay, line 1591: nit: "get" should be "gets" (or use "is never called")

PR #187, opened 2020-08-26
Add option to ignore unknown fields when parsing JSON.

bkuehnert:
Per proto spec, if the JSON parser encounters a field which is not
defined in the message, then an error should be thrown. This commit
implements this behavior and adds an option to ignore these unknown
fields when parsing.

PR #191, opened 2020-08-27
Update README w/ JSON documentation.

bkuehnert:
This should go in after #187, as it includes the `ignore-unknown-fields-p` option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+
 # cl-protobufs
 ![SBCL-Tests](https://github.com/qitab/cl-protobufs/workflows/SBCL-Tests/badge.svg?branch=master)
 ![CCL-Tests](https://github.com/qitab/cl-protobufs/workflows/CCL-Tests/badge.svg?branch=master)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -609,3 +609,28 @@ base-url/qualified-name.
 Creates an `Any` protobuf message given a protobuf `message` and a `base-url`.
 
 TODO: examples
+
+### JSON Mappings
```
cgay, line 613: I don't think this should be a child topic of `## Well Known Types`. How about moving it above, so it's a sibling of `## The cl-protobufs Package  >>  ### Serialization` instead?

cgay: It could also be considered a kind of serialization rather than a separate thing, in which case it could be included in the Serialization section, but I think that's a bit confusing. At least I usually think about serialization and "converting to json" as distinct, but that's probably a historical accident. Up to you.

README.md
```diff
@@ -609,3 +609,28 @@ base-url/qualified-name.
...
+
+The `cl-protobufs.json` package exports functions to convert between protobuf objects
+and [the canonical JSON encoding](https://developers.google.com/protocol-buffers/docs/proto3#json).
```
cgay, line 616: Some long lines here are going to cause lint errors internally, I think. Actually can't remember for sure if the linter cares about Markdown in `/third_party/`, but best to wrap the non-code lines at about 80 chars.

README.md
```diff
@@ -609,3 +609,28 @@ base-url/qualified-name.
...
+
+```lisp
+(defun print-json (message &key (indent 0) (stream \*standard-output\*) (camel-case-p t) numeric-enums-p))
+```
+
+Takes any protobuf message `message` and prints its JSON. The options are as follows:
+- `indent`: Indent the output by `indent` spaces. If `indent` is `nil`, then do not pretty-print the output.
+- `stream`: The lisp stream to output to.
+- `camel-case-p`: Print field names in camelCase. If `nil`, then print field names are they are defined in the .proto file.
```
cgay, line 625: "are they are" should be "as they are" (or "as they appear").

README.md
```diff
@@ -609,3 +609,28 @@ base-url/qualified-name.
...
+- `numeric-enums-p`: If true, print enum values by their number rather than their name.
+
+```lisp
+(defgeneric parse-json (type &key stream ignore-unknown-fields-p)
+```
+
+Parse a JSON encoding and return the parsed protobuf object. The options are as follows:
+- `type`: Either the lisp type or the `message-descriptor` of the object to parse.
```
cgay, line 633: "lisp" is generally capitalized: "Lisp"

README.md
```diff
@@ -609,3 +609,28 @@ base-url/qualified-name.
...
+- `stream`: The stream to parse from. By default, this is \*standard-input\*.
```
cgay, line 634: Pedantic: one doesn't parse from a stream, one reads from a stream (and then parses the data).

README.md
bkuehnert: I intended this to be a separate section entirely, that's a typo.

bkuehnert: How does it feel as it's own section entirely? I figure it should be a sibling of well known types, since JSON defines its own package.

cgay:
Doing this PR internally for copybara reasons.

Issue #199, opened 2020-08-29
abcl tests are using the quicklisp version of  cl-protobufs in testing


Slids:
Not sure why this is happening, but we can see running:
https://github.com/Slids/cl-protobufs/runs/1044380174?check_suite_focus=true
it's true.

easye:
@slids I fixed a regressin in ABCL that prevented Quicklisp from properly finding ASDF systems in the quicklisp/local-projects/ directory. 

<https://github.com/armedbear/abcl/commit/7789522b366b6a99eb19bc5652d61d8a0013c6a9>

You should be able to close this issue now.  Please lemme know if that is still not the case.

PR #207, opened 2020-09-09
ci: use --batch option for invoking ABCL

easye:
Use the `--batch` option to invoke ABCL, which causes the implementation to exit after processing and executing all other command line options.  

Although such an enhancement may shave a bit of time off executing the ABCL CI tests, the real purpose is to show that ABCL now passes tests after fixing the issue in <https://github.com/qitab/cl-protobufs/issues/199>


easye:
@googlebot I signed it!

Slids:
Thank you @easye !

PR #340, opened 2021-07-14
Remove calls to (quit) from abcl commands

cgay:
The `--batch` flag causes abcl to exit. The `(quit)` doesn't actually work there.


cgay:
Closing this since the ABCL tests are wedged at the moment.

PR #364, opened 2022-08-05
use enum name as constant prefix

jidibinlin:


Slids:
I like the idea, but you 
a) Need to sign the cla
b) fix the tests
c) We need to make an anouncement as this is a breaking change.

cgay (changes requested):
Please also update the tests. See the failed test run here: https://github.com/qitab/cl-protobufs/runs/7701713796?check_suite_focus=true

jidibinlin:
I will find a time to solve this issue

jidibinlin:
I have fixed the enum-mapping-test.lisp file。 But still can`t pass all the test。
image

cl-protobufs.asd
```diff
@@ -81,6 +81,8 @@ and functionality for working with them."
       :proto-pathname "google/protobuf/any.proto")
      (:protobuf-source-file "source_context"
       :proto-pathname "google/protobuf/source_context.proto")
+     (:protobuf-source-file "descriptor"
```
Slids, line 84: why add descriptor?

Slids: (not saying not to, just asking why)

cgay: It's already there on links 78-79.

PR #379, opened 2022-12-20
Use `float-features` to handle infinite value

bohonghuang:
Hi! Noticing `float-features` was added as a dependency recently, I made it used for handling `inf` and `nan`, which are valid values in protobuf.


Slids:
Thank you!
Please sign the cla and add tests.

bohonghuang:
@Slids All done. Please check them out.

Issue #402, opened 2023-04-26
error compiling cl-protobufs in order to install


vchudnov-g:
I'm having problems installing cl-protobufs. I get errors like 

```
g++ -std=c++14 -I/PATH/TO/MY/LOCAL/lisp/protobuf/include -I.   -c -o generator.o generator.cc
generator.cc: In member function ‘virtual bool google::protobuf::cl_protobufs::LispGenerator::Generate(const google::protobuf::FileDescriptor*, const std::string&, google::protobuf::compiler::OutputDirectory*, std::string*) const’:
generator.cc:62:5: error: ‘GOOGLE_CHECK’ was not declared in this scope
   62 |     GOOGLE_CHECK(annotations.SerializeToZeroCopyStream(meta.get()));
      |     ^~~~~~~~~~~~
generator.cc: In member function ‘virtual bool google::protobuf::cl_protobufs::LispGenerator::GenerateAll(const std::vector<const google::protobuf::FileDescriptor*>&, const std::string&, google::protobuf::compiler::GeneratorContext*, std::string*) const’:
generator.cc:135:5: error: ‘GOOGLE_CHECK’ was not declared in this scope
  135 |     GOOGLE_CHECK(annotations.SerializeToZeroCopyStream(meta.get()));
      |     ^~~~~~~~~~~~
make: *** [<builtin>: generator.o] Error 1
```

FWIW, I had to use different compilation steps for `protobuf` than what's included in your README, since `protobuf` seems to be currently set up for `bazel` or `cmake` builds. After trying various approaches, the following  are the steps that wound up successfully compiling protobuf and installing it (and its `include/`) to a local user directory for me. I don't get an error until I try to compile `cl-protobuf`:

```
export PROTOC_ROOT=/PATH/TO/MY/LOCAL/lisp/protobuf
mkdir -p /tmp/lisp/proto
pushd /tmp/lisp/proto

git clone --recursive https://github.com/google/protobuf
pushd protobuf
git submodule update --init --recursive

mkdir -p ${PROTOC_ROOT}
cmake . -DCMAKE_CXX_STANDARD=14 -DCMAKE_INSTALL_PREFIX=${PROTOC_ROOT}
cmake --build . --parallel 10

cmake --install .
path-prepend ${PROTOC_ROOT}/bin
popd # protobuf

git clone git@github.com:qitab/cl-protobufs.git

pushd cl-protobufs/protoc
make  #### ****  <-- error occurs here

[...]

PR #404, opened 2023-05-07
Update install guide to note issue #402

bkuehnert:
The cl-protobufs protoc plugin is not compatible with `protoc` versions v22+. This is because `protoc` now depends on abseil and removed implementations of utility functions that cl-protobufs relied on.

This update to the README does two things
1. Notes this problem, and updates the install steps to specify version 21.x.
2. Offers installation of `protoc` via package manager as an option, since it turns out that this works just fine.


mdbergmann:
Just a comment: building cl-protobufs from 'brew install protobuf@21' on macOS works fine. On Apple Silicon macOS the install path for `PROTOC_ROOT=/opt/homebrew/Cellar/protobuf@21/21.12`.

PR #434, opened 2024-03-29
Support newer versions of protobuf

Partmedia:
This library doesn't compile with newer versions of protobuf (see https://github.com/qitab/cl-protobufs/issues/402), because some functions have been moved/renamed.

After I completed this patch, I realized that this duplicates efforts that already exist:
- https://github.com/qitab/cl-protobufs/pull/426
- https://github.com/qitab/cl-protobufs/pull/420
- https://github.com/qitab/cl-protobufs/pull/401
- https://github.com/qitab/cl-protobufs/pull/397

But this is all tidy and in a smaller and more manageable diff. Ultimately because this is a Google library you might opt to close this PR and merge those others instead :)

fixes #402

PR #436, opened 2024-06-22
Update to use latest protobuf

michaeldelago:
This is based on the work in #434 by @Partmedia

This PR adds the missing `include` statements as well as a few tweaks to the Makefile to get everything working.

Changes to the Makefile

- use `-std=c++17` instead of `-std=c++14`. It wasn't working on my machine with 14
- use `pkg-config` to setup linking for absl and protobuf
  - absl apparently only supports Bazel and CMake. There's a BUILD file in there but I have no idea how to use bazel and the README.md states to use `make`

### DIFF
--- a/.github/workflows/ABCL-test.yml
+++ b/.github/workflows/ABCL-test.yml
@@ -39,7 +39,7 @@ jobs:
         run: echo "::add-path:$GITHUB_WORKSPACE/quicklisp/local-projects/cl-protobufs/protoc/"
 
       - name: Download and install protobuf
-        run: sudo apt install -y protobuf-compiler libprotobuf-dev libprotoc-dev
+        run: sudo apt install -y protobuf-compiler libprotobuf-dev libprotoc-dev libabsl-dev
 
       - name: Install protoc plug-in
         run: cd $GITHUB_WORKSPACE/quicklisp/local-projects/cl-protobufs/protoc && PROTOC_ROOT=/usr make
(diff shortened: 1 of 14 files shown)


Partmedia:
Thanks for improving on this. I have to remember to go back and fix the workflows on my PR :')

I might offer that the reason only c++17 works is that you may be linking to a system abseil library compiled with C++17 flags (the linked library is not cross-C++ runtime ABI compatible). That's why I also needed C++17 on my machine to get this to compile, but I opted not to include that in the PR.

Slids:
Fails CI (which should at least pass ccl and sbcl

michaeldelago:
CI fixed (maybe). They weren't installing absl in the setup steps.

> I might offer that the reason only c++17 works is that you may be linking to a system abseil library compiled with C++17 flags (the linked library is not cross-C++ runtime ABI compatible).

That makes sense. I'm using the current version of protobuf that's in the Arch Linux package repository, so I didn't compile it myself.

Slids:
CCL and SBCL are not yet passing CI

Slids:
So, the issue is the ubuntu repository used in ubuntu-latest doesn't have absl_absl_log in libabsl-dev.

Slids:
https://abseil.io/docs/cpp/tools/cmake-installs 
Will tell you how to build and install.

Installing is slow, not quite sure what to do about that, 
but I think it's the best way forward?

michaeldelago:
work picked up - unfortunately too distracted to give this as much time as I wanted

nicely, this is superseded by #438

